PR #2, opened 2016-03-09
temporary fixes

vepiteski:
I started from an old version and eventually duplicated previous changes. My actual version seems to work, but second derivatives in the new JuMP 0.12 are suspect. I still struggle to produce a small example of the trouble I get when solving iteratively a Brachistochrone model at iteration 107. This does not look as a NLPModels fault but rather comes from JuMP new AD implementation.

Once JuMP initializes to 0 the hv result in hprod, the temporary fix should be removed.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -3,7 +3,7 @@ module NLPModels
 using JuMP
 using MathProgBase
 
-require(Pkg.dir("MathProgBase", "src", "NLP", "NLP.jl"))
+include(Pkg.dir("MathProgBase", "src", "NLP", "NLP.jl"))
 using NLP  # Defines NLPModelMeta.
 
 export AbstractNLPModel, NLPModel,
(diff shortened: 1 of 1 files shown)


vepiteski:
The tests are broken because of comparisons with AMPL using a deprecated package (AMPLReader).

src/NLPModels.jl
```diff
@@ -275,6 +278,7 @@ end
 "Evaluate the product of the Lagrangian Hessian at `(x,y)` with the vector `v`."
 function hprod(nlp :: NLPModel, x :: Array{Float64}, y :: Array{Float64}, v :: Array{Float64})
   nlp.neval_hprod += 1
+  nlp.hv = zeros(size(x)) # JPD workaround
```
dpo: I'm not sure I understand this; wasn't `nlp.hv` already initialized to the vector of zeros at line 144?

src/NLPModels.jl
```diff
@@ -282,6 +286,7 @@ end
 "Evaluate the product of the Lagrangian Hessian at `(x,y)` with the vector `v` in place."
 function hprod!(nlp :: NLPModel, x :: Array{Float64}, y :: Array{Float64}, v :: Array{Float64}, hv :: Array{Float64})
   nlp.neval_hprod += 1
+  hv = zeros(size(x)) # JPD workaround
```
dpo: This creates a new vector of zeros. Wouldn't it be enough to fill the `hv` supplied by the user with zeros?

vepiteski:
This "fix" is no longer required since JuMP itself has been fixed and performs the initialization to 0.

PR #4, opened 2016-04-05
SimpleNLPModel

### LINKED ISSUE #3: SimpleNLPModel
Create a simple model with very little overhead. For instance
````
f(x) = dot(x,x)/2
g(x) = x
H(x) = eye(2)
x0 = [1.0; 2.0]
nlp = SimpleNLPModel(x0, obj=f, grad=g, hess=H)
````

### PR DESCRIPTION
**Pull request by abelsiqueira**
Closes #3

### CODE REVIEW

**Comment by dpo** [comment]
Great idea!

File: src/simple_model.jl [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+  f :: Function
+  g :: Function
+  g! :: Function
+  H :: Function
+  c :: Function
+  c! :: Function
+  J :: Function
+end
+
+function SimpleNLPModel(x0; y0 = [], lvar = [], uvar = [], lcon = [], ucon = [],
```
dpo: I think those should be typed.

src/simple_model.jl
```diff
@@ -0,0 +1,153 @@
...
+    nnzj = 0, nnzh = 0, obj = error, grad = error, grad! = error, hess =
```
dpo: At least `obj` should be mandatory I believe (there's no problem without an objective). The others should probably be `Function`s, and if not specified, they should probably be something like `() -> throw(error("not implemented"))`. So the arguments to the constructor should probably be of type `Union{Nothing, Function}`, and if `Nothing`, then we should set them to the default value. What do you think?

src/simple_model.jl
```diff
@@ -0,0 +1,153 @@
...
+    for j = 1:nlp.meta.nvar
+      for i = 1:nlp.meta.ncon
+        a = J[i,j]
+        if a != 0.0
+          push!(vals, a)
+          push!(rows, i)
+          push!(cols, j)
+        end
+      end
+    end
```
dpo: How about `findnz(sparse(J))`?

src/simple_model.jl
```diff
@@ -0,0 +1,153 @@
...
+    return rows, cols, vals
+  end
+end
+
+function jac(nlp :: SimpleNLPModel, x)
+  return nlp.J(x)
+end
+
+function jprod(nlp :: SimpleNLPModel, x, v)
+  return nlp.J(x)*v
```
dpo: A user should be allowed to provide `jprod` to the constructor. If `jac` is given but not `jprod`, then this is fine.

src/simple_model.jl
abelsiqueira: I can do this without the Union type, by defining
````
julia> NI(args...) = throw(error("Not"))
julia> function foo(;f = NI) println(f(0)) end
julia> foo(f = exp)
1.0
julia> foo()
ERROR: Not
````

src/simple_model.jl
abelsiqueira: Much better, thanks

src/simple_model.jl
abelsiqueira: Even better
````
julia> NI(n) = throw(error("$n not ..."))
julia> function foo(;f = (args...)->NI("f") println(f(0)) end
julia> foo(f = exp)
1.0
julia> foo()
ERROR: "f" not ...

abelsiqueira: Ok, there is a difference. On the first `f == NI`  is `true`, but not on the second. One way to circunvent is to `try f(x)` and catch the exception. Since `SimpleNLPModel` is not about efficiency, I'll go with this option.

dpo: `f` should still be typed as a `Function` though, shouldn't it?

abelsiqueira: Yes.
I'm creating a `NotImplementedError <: Exception` to handle this.

dpo:
If you want to rebase this, I'm now thinking we could pull it as-in and create a separate `SimpleADModel` that uses AD to compute derivatives. What do you think? Is it worth having both?

abelsiqueira:
I was thinking about that. I was gonna check how AD works first to see which is better.

dpo:
Ok. I can see how `SimpleNLPModel` could be useful if you need to call an external library to evaluate your objective, gradient, etc.

Issue #6, opened 2016-04-25
Consistency


abelsiqueira:
All models should

 - [x] return the same Hessian (only triangular);
 - [x] use the same Lagrangian (s*f(x) + sum y_ic_i + zl_i(x_i - l_i) + zu_i(u_i-x_i) );
 - [x] implement all functions (see #5);
 - [x] have the same meta;
 - [x] have the same counters.

There may be some exceptions. *SlackModel* for instance is consistent only for equality constrained problems.

abelsiqueira:
While trying to make meta the same for JuMP and AMPL, I noticed that NLP has a lot of information. Some very useful for our future selector, but very hard to implement for some models (CUTEst).
So, instead of having the same meta, I'll consider only a more reasonable subset of the fields.

Furthermore, do we want all those fields in meta? I think it was originally develop thinking of AMPL, but won't we focus on continuous NLP?

dpo:
There's no need to fill in all fields. AmplNLReader can be easily modified to fill them all in (it currently doesn't) but I don't see any harm in keeping them around. Because Julia doesn't support inheritance, it's not easy to extend a "meta" type.

abelsiqueira:
Understood. My concern here was with the bounds on the constraints, but I'll do something different now and still mantain some consistency: I'll compare c(x) - L and c(x) - U, instead of only c(x). So JuMP and AMPL should match.

dpo:
What was the concern with c(x)? Were you trying to determine the type of each constraint?

abelsiqueira:
I am trying to avoid specific cases like [this](https://github.com/JuliaOptimizers/NLPModels.jl/blob/develop/test/jump_vs_ampl.jl#L31-L36).

dpo:
It's important to make the difference between those types of constraints though, isn't it?

abelsiqueira:
Possibly, but I'm just avoiding it at the consistency tests.

abelsiqueira:
I'll have to make a little change on the ampl and jump models for counter consistency.
`jprod` is implemented by calling `jac` which increases the `jac` counter, so I'll have to stop calling `jac`, and manually call whatever is being called instead.

abelsiqueira:
Alternatively, I can sum `jac`, `jprod` and `jtprod` counters, which looks better than the alternative, and less code is changed.

dpo:
[...]

PR #11, opened 2016-06-07
Consistency start

abelsiqueira:
Changed `jump_vs_ampl` to a more general file, that hopefully can be used with other NLPs.

I'll work on more consistency checks, and further commits will make this PR very hard
to review, so I do this first, and smaller ones after.


abelsiqueira:
Relates to #6.

abelsiqueira:
By the way, it fails on 0.5 because of other packages.

dpo:
Looks great. Thank you!

Issue #5, opened 2016-04-25
Test all functions on all interfaces


abelsiqueira:
- Models
  - [x] SimpleNLPModel
  - [x] JuMPModel
  - [x] AMPLModel
  - [x] CUTEstModel
 - Functions (and inplace versions)
  - [x] obj(nlp, x)
  - [x] grad(nlp, x)
  - [x] hess(nlp, x)
  - [x] hess(nlp, x, y=y)
  - [x] hprod(nlp, x, v)
  - [x] hprod(nlp, x, v, y=y)
  - [x] cons(nlp, x)
  - [x] jac(nlp, x)
  - [x] jprod(nlp, x)
  - [x] jtprod(nlp, x)

dpo:
`JuMPModel` doesn't support Jacobian-vector products (yet). `AmplModel` doesn't either and likely won't.

abelsiqueira:
How do you wanna handle this jprod? Use a default build-the-matrix-then-prod implementation? Lose the requirement? PR them?

dpo:
I suppose the most robust temporary solution would be to evaluate the Jacobian and then multiply.

abelsiqueira:
I'm gonna have to change some sizes in the tests to have compatibility with CUTEst. I hope it's that simple.

abelsiqueira:
How do I convert mod to nl? Can I do it without the ampl executable?

dpo:
You need the ampl executable to generate a nl file. That's what it's for.

abelsiqueira:
Regarding the CUTEst compatibility, there will be a PR shortly, but there are some issues with the problems:
- GENROSE: Even with 100 variables (changing the SIF), there is considerable error. obj returns and error of 4, where f ~ 1e6. The maximum error expected is 1e6 * 1e-10 = 1e-4, and all other problems find it, so it is the way CUTEst compute it that gives such a large error.
- HS11: The constraint in CUTEst is the negative of the one on the rest.
- HS14: Something similar to HS11, haven't checked yet.

dpo:
If the problem formulations differ, we should make them what the official reference says (e.g., the HS book).

For genrose, what is the error that you are measuring?

abelsiqueira:
I'm measuring function values from other models and CUTEst.
The values are (for some random x0)
````
1.4995852340157446e6 # Other
1.4995553998517566e6 # CUTEst
````
[...]

PR #12, opened 2016-06-08
More consistency tests

abelsiqueira:
Regarding #5 and #6.

### DIFF
--- a/src/jump_model.jl
+++ b/src/jump_model.jl
@@ -198,6 +198,44 @@ function jac(nlp :: JuMPNLPModel, x :: Array{Float64})
   return sparse(jac_coord(nlp, x)..., nlp.meta.ncon, nlp.meta.nvar)
 end
 
+"""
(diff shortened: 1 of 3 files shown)


src/jump_model.jl
```diff
@@ -198,6 +198,44 @@ function jac(nlp :: JuMPNLPModel, x :: Array{Float64})
   return sparse(jac_coord(nlp, x)..., nlp.meta.ncon, nlp.meta.nvar)
 end
 
+"""
+Evaluate the Jacobian-vector product at `x`.
+Warning: Currently building the Jacobian for this.
+"""
+function jprod(nlp :: JuMPNLPModel, x :: Array{Float64}, v :: Array{Float64})
+  return jac(nlp, x) * v
```
dpo: As before, I think this should call `jprod!()`.

abelsiqueira:
Updated

PR #13, opened 2016-06-09
Implement SimpleNLPModel

abelsiqueira:
Includes consistency checks.

### DIFF
--- a/src/simple_model.jl
+++ b/src/simple_model.jl
@@ -1,0 +1,293 @@
+export SimpleNLPModel, obj, grad, grad!, cons, cons!, jac_coord, jac, jprod,
+       jprod!, jtprod, jtprod!, hess, hprod, hprod!
+
+type SimpleNLPModel <: AbstractNLPModel
+  meta :: NLPModelMeta
+
+  counters :: Counters
+
+  # Functions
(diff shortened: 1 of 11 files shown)


src/simple_model.jl
```diff
@@ -0,0 +1,272 @@
...
+  nvar = length(x0)
+  length(lvar) == 0 && (lvar = -Inf*ones(nvar))
+  length(uvar) == 0 && (uvar =  Inf*ones(nvar))
+  ncon = maximum([length(lcon); length(ucon); length(y0)])
+  if nnzh == 0
+    try
+      if ncon == 0
+        A = hess(x0)
+      else
+        A = hess(x0, y0)
```
dpo: Shouldn't you check that `y0` was specified here, and possible set it to a vector that's likely to bring out all constraint contributions? I would choose something like `y0 = [1, -1, 1, -1, ...]`.

abelsiqueira: True. I think I'm gonna create a fake y0 even if y0 is specified (it could be zeros).

abelsiqueira:
Updated

dpo:
Great, thanks!

Issue #15, opened 2016-07-03
Automatic differentiation package independent on JuMP


abelsiqueira:
Can it be done in an easy way without having to recreate JuMP?
- How easy?
- How much overhead?
- How usable?

abelsiqueira:
Closed by #17

PR #17, opened 2016-07-06
Change SimpleNLPModel to use automatic differentiation

abelsiqueira:
This makes it much easier to create a simple model.
Since SimpleNLPModel is not intented for speed, this is good for the user.

### DIFF
--- a/src/simple_model.jl
+++ b/src/simple_model.jl
@@ -95,27 +95,30 @@ function jtprod!(nlp :: SimpleNLPModel, x :: Vector, v :: Vector, Jtv :: Vector)
 end
 
 function hess(nlp :: SimpleNLPModel, x :: Vector; obj_weight = 1.0, y :: Vector = [])
-  Hx = ForwardDiff.hessian(nlp.f, x) * obj_weight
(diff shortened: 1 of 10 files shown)


abelsiqueira:
Closes #15

src/simple_model.jl
```diff
@@ -72,202 +43,79 @@ function SimpleNLPModel(x0::Vector, obj::Function; y0::Vector = [],
...
-    try
-      return nlp.Hprod!(x, w, v, Hv)
-    catch
-      Hv[:] = nlp.H(x,w) * v
-      return Hv
-    end
+    obj_weight = 1.0, y :: Vector = [])
+  Hv[:] = ForwardDiff.hessian(nlp.f, x) * v * obj_weight
+  for i = 1:length(y)
+    Hv[:] += ForwardDiff.hessian(x->nlp.c(x)[i], x) * v * y[i]
```
dpo: I think we could test whether `obj_weight` or any of the `y[i]` is zero here.

abelsiqueira:
Updated

dpo:
Looks great, thank you! Would you like to add the eval counters at this point?

abelsiqueira:
I will add them and the tests in another PR, after the CUTEst PR, otherwise it will be too large

PR #18, opened 2016-07-11
Add CUTEst to the consistency tests and fix other models

abelsiqueira:
Change the size of genrose to 500 to match default GENROSE.SIF.
Fix HS11 and HS14 to match the Hock-Schittkowski paper.
Remove GENROSE because it is too slow.
Add BROWNDEN.
Add git depth following

    https://github.com/travis-ci/travis-ci/issues/4942

Closes #5

PR #19, opened 2016-07-15
Test cutest with counters

abelsiqueira:
This is built on top of #18 
I can rebase later.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -31,8 +31,13 @@ function reset!(counters :: Counters)
   return counters
 end
 
+"Reset evaluation count in `nlp`"
+function reset!(nlp :: AbstractNLPModel)
+  reset!(nlp.counters)
+  return nlp
+end
+
 # Methods to be overridden in other packages.
-reset!(nlp :: AbstractNLPModel) = throw(NotImplementedError("reset!"))
 obj(nlp :: AbstractNLPModel, args...; kwargs...) =
   throw(NotImplementedError("obj"))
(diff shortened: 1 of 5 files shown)


src/NLPModels.jl
```diff
@@ -31,8 +31,13 @@ function reset!(counters :: Counters)
   return counters
 end
 
+"Reset evaluation count in `nlp`"
+function reset!(nlp :: AbstractNLPModel)
+  reset!(nlp.counters)
+  return nlp
+end
+
```
dpo: Doesn't this allow a user to define an object of type `AbstractNLPModel` and call `reset!()`, which would error out because there's no field named `counters`?

dpo: Maybe it's enough to parametrize? E.g.,
```
function reset!{T <: AbstractNLPModel}(nlp :: T)
  ...
end
```

dpo:
I'll wait for the rebase because it's difficult to tell what's changed exactly.

src/NLPModels.jl
abelsiqueira: No, because abstract types cannot be instantiated.

dpo: Ah, ok!

dpo:
Could you please rebase this?

abelsiqueira:
I rebased, but due to the last comment on #6, it's not completely done yet.

abelsiqueira:
Done

dpo:
Thank you.

PR #22, opened 2016-07-26
Define constraint in simple_model test

dpo:
The previous test did not define any constraint.

### DIFF
--- a/test/test_simple_model.jl
+++ b/test/test_simple_model.jl
@@ -4,7 +4,7 @@ function test_simple_model()
   nlp = SimpleNLPModel(x0, f)
 
   c(x) = [sum(x) - 1]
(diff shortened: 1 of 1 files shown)


abelsiqueira:
Nice catch, thanks. But I just found out that I don't have write access to this repository.

dpo:
I'm not sure why TravisCI isn't kicking in.

abelsiqueira:
The osx release had a [strange error downloading brew](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/147549746#L154). I tried restarting but it said I couldn't.

dpo:
Clearly a Travis  bug. Let's pull this since the other three checked out.

PR #23, opened 2016-08-19
Documentation

abelsiqueira:
This documentation uses Documenter.jl.
To publish you need to first do the steps [here](https://juliadocs.github.io/Documenter.jl/latest/man/hosting.html). More specifically, create a ssh key to store in github and then commit the encripted key. Also, the documentation is only created if a specific branch is pushed (currently develop) or in a tag.

The current documentation was pushed [here](http://abelsiqueira.github.io/NLPModels.jl/latest/).

Summarizing:
- Accept this
- Create ssh key and commit the encripted key (verify the [instructions](https://juliadocs.github.io/Documenter.jl/latest))
- Push to develop

If you give me admin access to this repository, I can do it from here.


dpo:
Many thanks for this!!!

Is MathJax supported? If so I think it would help to write the problem in math form and refer to the objective and constraints in the documentation.

abelsiqueira:
Updated the docs with the suggestion.

dpo:
Thanks! Will we need to push updated doc manually every time or can Travis do it?

abelsiqueira:
In theory, after correcting the .documenter.enc and the keys on travis and github it should deploy for the develop branch as the latest doc and for the releases also. I don't remember now about the stable docs. Maybe it's the master, maybe it's' the latest release.

dpo:
I gave you admin access. Many thanks for this!

abelsiqueira:
Dropping this because of #25

PR #25, opened 2016-09-01
Documentation

abelsiqueira:
Tested inside the organization, and now with badges.

http://juliasmoothoptimizers.github.io/NLPModels.jl/latest/


dpo:
💯

dpo:
I'm thinking that the docs could specify clearly our convention about signs in the Lagrangian.

abelsiqueira:
Convention added

abelsiqueira:
@dpo, the fail here is only on Coverage.
Also, this introduced a change on SimpleNLPModel that I'm using on the tests on TRON.

dpo:
Is this ready to go?

abelsiqueira:
Yes

sent from cell

PR #26, opened 2016-09-03
model names

dpo:


abelsiqueira:
This is very cool. A good addition to OptimizationProblems.
Let's wait for the docs to see if they'll need to be changed, and if the doc creation/push is working correctly.

abelsiqueira:
Thanks. Merged manually with an additional commit for documentation, which was pushed automatically.

PR #28, opened 2016-09-11
Quasi-Newton Models

dpo:
A simple initial design for quasi-Newton models. Example
```julia
julia> using NLPModels
julia> model = SimpleNLPModel(x -> dot(x, x), zeros(2))
julia> qnmodel = LBFGSModel(model)  # or LSR1Model; optional memory argument
```
Differences with other types of models:
* `hess_coord()` is not implemented (the Hessian is typically dense)
* `hess()` ~~returns a `LinearOperator`~~ is not implemented
* `hprod()` applies the linear operator

I'll add tests when we agree on the design.

@abelsiqueira

### DIFF
--- a/src/qn_model.jl
+++ b/src/qn_model.jl
@@ -1,0 +1,64 @@
+export QuasiNewtonModel, LBFGSModel, LSR1Model,
+       reset!,
+       obj, grad, grad!,
+       cons, cons!, jac_coord, jac, jprod, jprod!, jtprod, jtprod!,
+       hprod, hess_op,
+       push!
+
+abstract QuasiNewtonModel <: AbstractNLPModel
+
+type LBFGSModel <: QuasiNewtonModel
+  meta :: NLPModelMeta
+  model :: AbstractNLPModel
+  op :: LBFGSOperator
+end
+
+type LSR1Model <: QuasiNewtonModel
+  meta :: NLPModelMeta
+  model :: AbstractNLPModel
+  op :: LSR1Operator
+end
+
+"Construct a `LBFGSModel` from another type of model."
+function LBFGSModel(nlp :: AbstractNLPModel; memory :: Int=5)
+  op = LBFGSOperator(nlp.meta.nvar, memory)
+  return LBFGSModel(nlp.meta, nlp, op)
+end
+
+"Construct a `LSR1Model` from another type of nlp."
+function LSR1Model(nlp :: AbstractNLPModel; memory :: Int=5)
+  op = LSR1Operator(nlp.meta.nvar, memory)
+  return LSR1Model(nlp.meta, nlp, op)
+end
+
+# retrieve counters from underlying model
+for counter in fieldnames(Counters)
+  @eval begin
+    $counter(nlp :: QuasiNewtonModel) = $counter(nlp.model)
+    export $counter
+  end
+end
+
+function reset!(nlp :: QuasiNewtonModel)
+  reset!(nlp.model.counters)
+  reset!(nlp.op)
+  return nlp
+end
+
+# the following methods are not affected by the Hessian approximation
+for meth in (:obj, :grad, :grad!, :cons, :cons!, :jac_coord, :jac, :jprod, :jprod!, :jtprod, :jtprod!)
+  @eval begin
+    $meth(nlp :: QuasiNewtonModel, args...) = $meth(nlp.model, args...)
+  end
+end
+
+# the following methods are affected by the Hessian approximation
+hess_op(nlp :: QuasiNewtonModel, x :: Vector{Float64}; kwargs...) = nlp.op
+hprod(nlp :: QuasiNewtonModel, x :: Vector{Float64}, v :: Vector{Float64}; kwargs...) = nlp.op * v
+
+function push!(nlp :: QuasiNewtonModel, args...)
+	push!(nlp.op, args...)
+	return nlp
+end
+
+# not implemented: hess_coord, hess, hprod!
--- a/src/qn_model.jl
+++ b/src/qn_model.jl
@@ -41,10 +41,14 @@ function reset!(nlp :: QuasiNewtonModel)
 end
 
 # the following methods are not affected by the Hessian approximation
-for meth in (:obj, :grad, :grad!, :cons, :cons!, :jac_coord, :jac, :jprod, :jprod!, :jtprod, :jtprod!)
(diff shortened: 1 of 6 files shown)


abelsiqueira:
Very good. I think the update with a separate function is best. Something like `update!(nlp)`. It should be implemented by default for an AbstractNLPModel so that calling it for a non-qn model won't raise anything.
Another thing, I think `hess` should also be undefined, because it should returns a triangular matrix.
Also, `reset!` should reset the operator too.

dpo:
Ok, thanks. A few points to sort out:

* I suggest that all `AbstractNLPModel`s have a new method, say `hess_op()` that returns the Hessian as a linear operator (based on `hprod()`). `QuasiNewtonModel`s will implement `hess_op()` and not `hess()`. Does that make sense?
* does a `QuasiNewtonModel` really need its own `counters` field? I'm thinking not, but then we need an API to access the counters across model types.
* ok for an `update!()` method.

abelsiqueira:
> I suggest that all AbstractNLPModels have a new method, say hess_op() that returns the Hessian as a linear operator (based on hprod()). QuasiNewtonModels will implement hess_op() and not hess(). Does that make sense?

Yes, and strongly agree with `hess_op`.
> does a QuasiNewtonModel really need its own counters field? I'm thinking not, but then we need an API to access the counters across model types.

Does `hprod` on a QNM increases the `hprod` counter of the inner model?

dpo:
> Does hprod on a QNM increases the hprod counter of the inner model?

No because an actual hprod was never computed. But perhaps `LinearOperators` should also contain a counter?!

abelsiqueira:
Don't know if a counter in LO would be beneficial for everyone. But a counter `Bprod` inside the QuasiNewtonModel might. Either way, that's an additional counter to worry about.
Otherwise, I might say the `counters(nlp)` returning a correct counter should be enough. It still might for most cases, I think.

Possibly, someone actually interested in `Bprod` will be careful enough to only look for Bprods inside QNMs?

dpo:
> Possibly, someone actually interested in Bprod will be careful enough to only look for Bprods inside QNMs?

Possibly, but they would only be counting a handful of dot products per hprod.

abelsiqueira:
So no `Bprod`? In that case, I mantain `counters(nlp)`.

dpo:
`SlackModel`s don't have a `counter` either because the underlying nlp is already counting. I feel we should have an API of the form
```julia
neval_obj(nlp :: AbstractNLPModel) = nlp.counters.neval_obj
# etc.
```
so that we can define
```julia
neval_obj(nlp :: SlackModel) = neval_obj(nlp.model)
neval_obj(nlp :: QuasiNewtonModel) = neval_obj(nlp.model)
# etc.
```
What do you think?

abelsiqueira:
That's even better.

dpo:
At this point, I'm too tired to figure out where this error is coming from...
```
Checking LBFGS formulation of hs10     	ERROR: LoadError: LoadError: push! not implemented
```

src/qn_model.jl
```diff
@@ -0,0 +1,63 @@
...
+for counter in fieldnames(Counters)
+  @eval begin
+    $counter(nlp :: QuasiNewtonModel) = $counter(nlp.model)
+    export $counter
+  end
+end
+
+function reset!(nlp :: QuasiNewtonModel)
+  reset!(nlp.model.counters)
+  return nlp
```
abelsiqueira: Can we reset the Operator here?

dpo: Great question! That's not currently possible but it's necessary!

abelsiqueira: Done in JuliaSmoothOptimizers/LinearOperators.jl/pull/16.
`LinearOperators.reset!(nlp.op)` should be enough.

test/test_qn_model.jl
```diff
@@ -0,0 +1,59 @@
...
+  v = rand(qnmodel.meta.nvar)
+  u = rand(qnmodel.meta.ncon)
+
+  @assert all(jprod(model, x, v) == jprod(qnmodel, x, v))
+  @assert neval_jprod(model) == 2
+
+  @assert all(jtprod(model, x, u) == jtprod(qnmodel, x, u))
+  @assert neval_jtprod(model) == 2
+
+  H = hess_op(model, x)
```
abelsiqueira: This should be `qnmodel`

test/test_qn_model.jl
```diff
@@ -0,0 +1,59 @@
...
+  @assert typeof(H) <: LinearOperators.AbstractLinearOperator
+  @assert size(H) == (model.meta.nvar, model.meta.nvar)
+  @assert all(H * v == hprod(model, x, v))
```
abelsiqueira: This should also be `qnmodel` and `==` verifies all components. `all` is not necessary.

dpo:
Updated.

src/qn_model.jl
```diff
@@ -0,0 +1,64 @@
...
+for counter in fieldnames(Counters)
+  @eval begin
+    $counter(nlp :: QuasiNewtonModel) = $counter(nlp.model)
+    export $counter
+  end
+end
+
+function reset!(nlp :: QuasiNewtonModel)
+  reset!(nlp.model.counters)
+  reset!(nlp.op)
```
abelsiqueira: `LinearOperators.reset!`

dpo: This is mysterious to me. There's a `using LinearOperators` at the top of `NLPModels.jl` (which `include`s `qn_model.jl`). Why doesn't multiple dispatch do its job here?

dpo: We have to explicitly `import` the methods we override.

abelsiqueira: Now it's failing because `LinearOperators.reset!` is in the latest develop.

dpo: Right so we need a new release of LinearOperators. Do you see anything else you'd like to get in there very soon?

abelsiqueira: Nothing in mind.

dpo:
Hm. I wonder why Travis and Appveyor aren't kicking in.

dpo:
How can the same build options behave differently on different repos?!

dpo:
Seems like a TravisCI bug.

dpo:
I think this is happy now.

abelsiqueira:
Thank you

PR #30, opened 2016-09-11
Counters API

dpo:
@abelsiqueira


dpo:
ps: I propose to use "squash and merge" when a PR consists in a single commit, and "create merge commit" when there are multiple commits.

abelsiqueira:
I'll try to keep that in mind

PR #31, opened 2016-09-12
Test jprod and jtprod in SlackModel

dpo:


dpo:
Testing uncovered a number of bugs here!

src/jump_model.jl
```diff
@@ -197,7 +197,7 @@ function jtprod!(nlp :: JuMPNLPModel,
                 Jtv :: Array{Float64})
   nlp.counters.neval_jac -= 1
   nlp.counters.neval_jtprod += 1
-  Jtv[:] = jac(nlp, x)' * v
+  Jtv[1:nlp.meta.nvar] = jac(nlp, x)' * v
```
dpo: I'm not sure why this doesn't bite us with `AmplModel`s.

dpo: Oh well, it probably does. It's just that `SlackModel(AmplModel)` isn't tested.

abelsiqueira:
Yikes, I completely missed [this line](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/develop/src/slack_model.jl#L45). Slack model is completely untested in terms of consistency. I'll merge this and see what I can do.

dpo:
Thanks!

Issue #32, opened 2016-09-12
Test SlackModel


abelsiqueira:
As uncovered in #31, slack is not being tested in of consistency. Trying to fix it is showing many small details, which I'm gonna post on other issues for resolution.

For now, what I'm trying to do to test the SlackModels is
```julia
# If there are inequalities, test the SlackModels of each of these models
  meta = nlp_simple.meta
  if meta.ncon > 0 && meta.ncon != length(meta.jfix)
    @printf("Checking slack variation of problem %-15s\t", problem_s)
    nvar = meta.nvar
    ncon = meta.ncon
    I = [meta.jlow; meta.jupp; meta.jrng]
    nineq = length(I)
    f(x) = nlp_simple.f(x[1:nvar])
    c(x) = begin
      cx = nlp_simple.c(x[1:nvar])
      for i = 1:nineq
        cx[I[i]] -= x[nvar+i]
      end
      return cx
    end
    P = ExtensionOperator(meta.jfix, ncon)
    slp = SimpleNLPModel(f, [meta.x0; zeros(nineq)], c=c,
        lvar=[meta.lvar; meta.lcon[I]], uvar=[meta.uvar; meta.ucon[I]],
        lcon=P*meta.lcon[meta.jfix], ucon=P*meta.ucon[meta.jfix])

    slack_nlps = [SlackModel(nlp) for nlp in nlps]
    push!(slack_nlps, slp)
    #consistent_meta(slack_nlps)
    consistent_functions(slack_nlps)
    @printf("✓\n")
  end
```
This creates a SimpleNLPModel with slacks using the other simple model functions (less intrusive), and then it tests everything (the functions only, at first).

Any drawbacks here?

dpo:
I think you could just test that `meta.ncon > length(meta.jfix)`.

Since `test_slack_model.jl` already tests that a slack formulation is correct, why not simply do
```julia
nlps = ...  # as before
consistent_functions(nlps)
@printf("✓\n")
slack_nlps = [SlackModel(nlp) for nlp in nlps]
consistent_functions(slack_nlps)
@printf("✓\n")
```

abelsiqueira:
I'll test that

abelsiqueira:
Done

PR #35, opened 2016-09-13
SlackNLPModel consistency and hess_coord consistency

abelsiqueira:


abelsiqueira:
Closes #32

src/simple_model.jl
```diff
@@ -96,7 +96,14 @@ end
...
 function grad!(nlp :: SimpleNLPModel, x :: Vector, g :: Vector)
   nlp.counters.neval_grad += 1
-  return ForwardDiff.gradient!(g, nlp.f, x)
+  if length(x) == length(g)
+    return ForwardDiff.gradient!(g, nlp.f, x)
+  else
+    gx = zeros(length(x))
+    ForwardDiff.gradient!(gx, nlp.f, x)
+    g[1:length(x)] = gx
+    return g
```
dpo: I wonder if this can be resolved with array views?!

abelsiqueira: Thanks, I updated it with SubArray (ArrayView's sugestion for 0.4)

src/simple_model.jl
```diff
@@ -99,9 +99,8 @@ function grad!(nlp :: SimpleNLPModel, x :: Vector, g :: Vector)
   if length(x) == length(g)
     return ForwardDiff.gradient!(g, nlp.f, x)
   else
-    gx = zeros(length(x))
+    gx = sub(g, 1:length(x))
     ForwardDiff.gradient!(gx, nlp.f, x)
-    g[1:length(x)] = gx
```
dpo: Is the `if ... else` still needed, or can we always do `ForwardDiff.gradient!(sub(g, 1:length(x)), nlp.f, x)`?

abelsiqueira: You're right, the sub call is inexpensive.

src/simple_model.jl
```diff
@@ -128,7 +128,7 @@ end
 
 function jprod!(nlp :: SimpleNLPModel, x :: Vector, v :: Vector, Jv :: Vector)
   nlp.counters.neval_jprod += 1
-  Jv[:] = ForwardDiff.jacobian(nlp.c, x) * v
+  Jv[1:nlp.meta.ncon] = ForwardDiff.jacobian(nlp.c, x) * v
```
dpo: So now I'm wondering if this shouldn't remain `Jv[:]` and if the caller should use `sub(...)`. What do you think?

abelsiqueira: Speedwise is the same, and this (should) break less, so maybe this way is better.
If we change this back, we have to change Ampl back too.

dpo: Yes, and JuMP. I guess this is safe, so let's leave it.

src/simple_model.jl
```diff
@@ -155,6 +155,17 @@ function hess(nlp :: SimpleNLPModel, x :: Vector; obj_weight = 1.0, y :: Vector
...
 
+function hess_coord(nlp :: SimpleNLPModel, x :: Vector; obj_weight = 1.0, y :: Vector = [])
+  Hx = sparse(hess(nlp, x, obj_weight=obj_weight, y=y))
+  rows = rowvals(Hx)
+  vals = nonzeros(Hx)
+  cols = Int[]
+  for i = 1:nlp.meta.nvar
+    append!(cols, i*ones(Int, length(nzrange(Hx,i))))
+  end
+  return rows, cols, vals
```
dpo: Doesn't `findnz()` work here?

abelsiqueira: Yes, thank you

dpo: It's too bad ForwardDiff only does dense Hessians. ReverseDiffSparse seems quite involved.

abelsiqueira: Yes, I hope to take a look at it, eventually. And the Autograd.jl too.

dpo:
👍

PR #42, opened 2016-09-22
Update to julia-0.5

abelsiqueira:
Stop the warnings on 0.5


abelsiqueira:
JuliaSmoothOptimizers/AmplNLReader.jl/pull/27, JuliaSmoothOptimizers/CUTEst.jl/pull/79 and JuliaSmoothOptimizers/NLPModels.jl/pull/42 have circular dependency on the tests, so it'll still show warnings.

abelsiqueira:
Heads up, Travis release is not 0.5 yet.

test/consistency.jl
```diff
@@ -201,13 +201,12 @@ function consistency(problem :: Symbol; nloops=100, rtol=1.0e-8)
...
   problem_f = eval(problem)
   nlp_ampl = AmplModel(joinpath(path, "$problem_s.nl"))
   nlp_jump = JuMPNLPModel(problem_f())
   nlp_simple = eval(parse("$(problem)_simple"))()
   nlps = [nlp_ampl; nlp_jump; nlp_simple]
-  @unix_only begin
+  @static if is_unix()
     nlp_cutest = CUTEstModel(uppercase(problem_s))
     push!(nlps, nlp_cutest)
   end
```
dpo: To eliminate the circular dependencies, we should probably eliminate AmplModel and CUTEstModel from here. It's safe to compare a CUTEstModel with a corresponding JuMPNLPModel in CUTEst.jl and an AmplModel with a corresponding JuMPNLPModel in AmplNLReader since those packages rely on NLPModels.jl. What do you think?

abelsiqueira: Agreed

abelsiqueira:
Updated. I also removed `JuMP` from `REQUIRE` because the code uses it optionally. The same will be done for `ForwardDiff` in `ADNLPModel`

abelsiqueira:
I broke the consistency function in two so I can reuse it on CUTEst and Ampl

dpo:
Did you mean to add an AUTOMAT.d and OUTSDIF.d?

abelsiqueira:
No. Removed.

dpo:
There are conflicts here now. Would you mind rebasing?

test/runtests.jl
```diff
@@ -18,7 +18,6 @@ for meth in filter(f -> isa(eval(f), Function), names(NLPModels))
   @test_throws(NotImplementedError, meth(model))
 end
 
-include("genrose.jl")
```
dpo: It's not clear to me what lines 12-19 are testing. `model` is defined as a `SimpleNLPModel`, so `obj`, etc., **are** defined and implemented, but the test just isn't using the appropriate calling sequence. `obj(model, x)` works. `obj(model)` doesn't exist but is made to throw `NotImplementedError` because `obj` is declared by default as `obj(nlp :: AbstractNLPModel, args...; kwargs...)`.

To me that's misleading because `obj` should always be `obj(nlp, x)` and never anything else.

I propose to redefine `obj()` as `obj(nlp :: AbstractNLPModel, x :: AbstractVector)`, and similarly for the other methods. Then the test could consist in something like
```
type DummyNLPModel <: AbstractNLPModel
end

dummy = DummyNLPModel()
@test_throws(NotImplementedError, obj(dummy, ones(2))
```
What do you think?

abelsiqueira:
Thank you, rebased.

test/runtests.jl
abelsiqueira: That makes much more sense. I think those are testing only if the appropriate exception is being thrown, but testing specific code is better.
Can I do it in another PR?

dpo: Yes it would make more sense.

PR #43, opened 2016-09-23
Factor out JuMP

dpo:
`JuMPNLPModel` is now called `MathProgNLPModel` since JuMP is not strictly required. It is still possible to create a `MathProgNLPModel` from a `JuMP.Model`, and that's indeed the current way to create one.

The IPOPT test now seems irrelevant to this package.


dpo:
There are a few conflicts with #42 here.

abelsiqueira:
This looks good to me. Can you rebase?

dpo:
Rebased.

abelsiqueira:
Thanks

PR #48, opened 2016-09-25
Optional check_derivatives on consistent_nlps and qualify reset! as NLPModels.reset!

abelsiqueira:
I got errors on `reset!` that weren't happening before AFAIR. Don't know why, but they are qualified now.

https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/pull/29

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -47,7 +47,7 @@ end
 obj(model, model.meta.x0)
 @assert neval_obj(model) == 1
 
-reset!(model)
+NLPModels.reset!(model)
 @assert neval_obj(model) == 0
 
 @test_throws(NotImplementedError, jth_con(model, model.meta.x0, 1))
--- a/test/consistency.jl
+++ b/test/consistency.jl
@@ -195,23 +195,25 @@ function consistent_functions(nlps; nloops=100, rtol=1.0e-8)
 
 end
 
-function consistent_nlps(nlps; nloops=100, rtol=1.0e-8)
+function consistent_nlps(nlps; nloops=100, rtol=1.0e-8, check_derivatives=true)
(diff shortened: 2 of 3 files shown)


dpo:
I'm not getting any error. What's the error message?

Is the optional `check_derivative` still useful if we have safe mode for `AmplModel`s?

abelsiqueira:
No, safe mode should be enough.

Error is
```
WARNING: both LinearOperators and NLPModels export "reset!"; uses of it in module Main must be qualified
ERROR: LoadError: UndefVarError: reset! not defined
 in include_from_node1(::String) at loading.jl:488
 in process_options(::Base.JLOptions) at client.jl:262
 in _start() at client.jl:318
while loading /home/abel/.julia/v0.4/NLPModels/test/runtests.jl, in expression starting on line 50
```
Both NLPModels and LinearOperators are on `develop`. The path is just a symlink. The error happens on 0.4, 0.5 and nightly.

test/runtests.jl
```diff
@@ -47,7 +47,7 @@ end
 obj(model, model.meta.x0)
 @assert neval_obj(model) == 1
 
-reset!(model)
+NLPModels.reset!(model)
```
dpo: This probably means we should have `import LinearOperators.reset!` before defining `reset!` in `NLPModels`?!

dpo:
I see. So this is with `LinearOperators`/`develop`?!

abelsiqueira:
Yes, I've been using LO/develop since `reset!`. I don't remember how much it's been dependent on it.
And you're right, importing solves it. I forgot we haven't merged #28, in which we import it.

I'm closing this.

PR #50, opened 2016-09-27
Update printing method to work with io

abelsiqueira:
Removes `1x3 Array`.
Uses `...` for larger than 5 components.
Accepts kw verbose to show full array.
Works with Documenter.jl


src/nlp_types.jl
```diff
@@ -168,36 +168,57 @@ end
...
+    print(io, meta; verbose=false)
+
+Prints meta information - x0, nvar, ncon, etc.
+If `verbose` is `false`, then vectors with more than 5 componens show only the
+first 4 and the last, like
+
+    1.0 2.0 3.0 4.0 ... 10.0
+"""
+function print(io :: IO, nlp :: NLPModelMeta; verbose=false)
+  nlp.minimize ? @printf(io, "Minimization ") : @printf(io, "Maximization ")
```
dpo: I think we could get rid of `verbose`, couldn't we? If a user doesn't want details, they should call `show()`.

abelsiqueira: :+1:

abelsiqueira:
Updated

dpo:
Would you mind adding a test printing and showing a problem with more than 5 variables and constraints?

abelsiqueira:
Done

dpo:
Thanks. I don't get it, it's still printing all the components: https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/163706568#L3059

abelsiqueira:
I understood you wanted to get rid of verbose and print all components every time.
Should I get rid of verbose and use the `...`, then?

dpo:
Sorry, yes, I think it's good to get rid of `verbose`, but I like your update that only prints a few components when there are more than 5. It's too bad Julia doesn't do that already.

abelsiqueira:
Done

dpo:
Thank you!

PR #54, opened 2016-10-02
Add Pkg.add on documentation

### LINKED ISSUE #49: Missing Pkg.checkout on page

abelsiqueira:
Closes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ develop
 
 **Documentation:**
 develop
-[![](https://img.shields.io/badge/docs-latest-blue.svg)](https://JuliaSmoothOptimizers.github.io/NLPModels.jl/latest)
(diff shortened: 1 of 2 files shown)


dpo:
In fact, if we merge develop into master, we don't need to do this, do we?

abelsiqueira:
No. Also, when releasing, it would change to `add` only.

dpo:
Should that information be in the README instead of the docs?

abelsiqueira:
I think it should be on both.

dpo:
Ok. do you want to update this PR so it says `add` and also update the README in prevision of a release?

abelsiqueira:
Yes, I'll do that

abelsiqueira:
Updated. Included some things on the docs that were on the README.

dpo:
👍 Let me know if this is ready.

abelsiqueira:
Ready

PR #55, opened 2016-10-04
Types of problems

### LINKED ISSUE #51: Simple queries on NLPModelMeta and AbstractNLPModel for type of problem
Is it useful enough? All oneliners (so far)

- is_free
- is_unc
- all_cons_linear
- all_cons_equality
- all_cons_inequality

abelsiqueira:
Closes #51

### DIFF
--- a/src/nlp_types.jl
+++ b/src/nlp_types.jl
@@ -223,7 +223,6 @@ end
 Returns whether the problem has bounds on the variables.
 """
 has_bounds(meta::NLPModelMeta) = length(meta.ifree) < meta.nvar
-has_bounds(nlp::AbstractNLPModel) = has_bounds(nlp.meta)
 
 """
     bound_constrained(nlp)
@@ -232,7 +231,6 @@ has_bounds(nlp::AbstractNLPModel) = has_bounds(nlp.meta)
 Returns whether the problem has bounds on the variables and no other constraints.
 """
 bound_constrained(meta::NLPModelMeta) = meta.ncon == 0 && has_bounds(meta)
-bound_constrained(nlp::AbstractNLPModel) = bound_constrained(nlp.meta)
 
 """
     unconstrained(nlp)
@@ -241,7 +239,6 @@ bound_constrained(nlp::AbstractNLPModel) = bound_constrained(nlp.meta)
 Returns whether the problem in unconstrained.
 """
 unconstrained(meta::NLPModelMeta) = meta.ncon == 0 && !has_bounds(meta)
-unconstrained(nlp::AbstractNLPModel) = unconstrained(nlp.meta)
 
 """
     linearly_constrained(nlp)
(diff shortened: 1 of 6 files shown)


src/nlp_types.jl
```diff
@@ -212,3 +215,50 @@ function print(io :: IO, nlp :: NLPModelMeta)
...
 end
+
+"""
+    bound_constrained(nlp)
+    bound_constrained(meta)
+
+Returns whether the problem has bounds on the variables.
+"""
+bound_constrained(meta::NLPModelMeta) = length(meta.ifree) < meta.nvar
+bound_constrained(nlp::AbstractNLPModel) = bound_constrained(nlp.meta)
```
dpo: Usually when we say a problem is bound constrained, we mean that it has *only* bound constraints. Here, the problem could have other contraints, right?

src/nlp_types.jl
```diff
@@ -212,3 +215,50 @@ function print(io :: IO, nlp :: NLPModelMeta)
...
+
+"""
+    unconstrained(nlp)
+    unconstrained(meta)
+
+Returns whether the problem in unconstrained.
+"""
+unconstrained(meta::NLPModelMeta) = meta.ncon == 0
+unconstrained(nlp::AbstractNLPModel) = unconstrained(nlp.meta)
```
dpo: I think this tests whether a problem is bound constrained.

src/nlp_types.jl
abelsiqueira: I'll change `bound_constrained` to that.
Can I add `has_bounds`?

src/nlp_types.jl
abelsiqueira: I'll change this to check for no bounds too.

src/nlp_types.jl
dpo: Of course! Good idea.

src/nlp_types.jl
```diff
@@ -212,3 +215,59 @@ function print(io :: IO, nlp :: NLPModelMeta)
...
 end
+
+"""
+    has_bounds(nlp)
+    has_bounds(meta)
+
+Returns whether the problem has bounds on the variables.
+"""
+has_bounds(meta::NLPModelMeta) = length(meta.ifree) < meta.nvar
+has_bounds(nlp::AbstractNLPModel) = has_bounds(nlp.meta)
```
dpo: If you wanted, and in case we add more such methods, those could all be defined at once:
```julia
for meth in (:has_bounds, :bound_constrained, ...)
  @eval $meth(nlp::AbstractNLPModel) = $meth(nlp.meta)
end
```

dpo:
Ready?

abelsiqueira:
Ready

dpo:
Thank you!

PR #64, opened 2016-10-22
mpb_model: specify desired features

### LINKED ISSUE #63: Add options for what is required of a MathProgNLPModel
Currently, `MathProgNLPModel(problem)` [sets up the structure for the Hessian and the expression graph](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/src/mpb_model.jl#L46), which takes large amounts of time for some problems. There should be an option to select what should be generated.

For example,

``` julia
using OptimizationProblems
prob = penalty2(5000)
model = MathProgNLPModel(prob)
```

takes several minutes, but if we request only

``` julia
MathProgBase.initialize(eval, [:Grad, :Jac, :HessVec])
```

then,

``` julia
julia> model = MathProgNLPModel(prob);

julia> @time model = MathProgNLPModel(prob);
  0.017955 seconds (503 allocations: 19.509 MB, 18.29% gc time)
```

dpo:
Closes #63

### DIFF
--- a/src/mpb_model.jl
+++ b/src/mpb_model.jl
@@ -40,10 +40,7 @@ function MathProgBase.loadproblem!(m :: MathProgModel,
                                    l, u, lb, ub,
                                    sense,
                                    eval :: MathProgBase.AbstractNLPEvaluator)
-
-  # TODO: :JacVec is not yet available.
-  # [:Grad, :Jac, :JacVec, :Hess, :HessVec, :ExprGraph]
-  MathProgBase.initialize(eval, [:Grad, :Jac, :Hess, :HessVec, :ExprGraph])
+  # eval is initialized in the MathProgNLPModel constructor
   m.numVar = numVar
   m.numConstr = numConstr
   m.x = zeros(numVar)
@@ -76,7 +73,12 @@ type MathProgNLPModel <: AbstractNLPModel
 end
 
 "Construct a `MathProgNLPModel` from a `MathProgModel`."
-function MathProgNLPModel(mpmodel :: MathProgModel; name :: String="Generic")
+function MathProgNLPModel(mpmodel :: MathProgModel;
+                          name :: String="Generic",
+                          features :: Vector{Symbol}=[:Grad, :Jac, :Hess, :HessVec, :ExprGraph])
+
+  # TODO: :JacVec is not yet available.
+  MathProgBase.initialize(mpmodel.eval, features)
 
   nvar = mpmodel.numVar
   lvar = mpmodel.lvar
@@ -89,10 +91,22 @@ function MathProgNLPModel(mpmodel :: MathProgModel; name :: String="Generic")
   lcon = mpmodel.lcon
   ucon = mpmodel.ucon
 
-  jrows, jcols = MathProgBase.jac_structure(mpmodel.eval)
-  hrows, hcols = MathProgBase.hesslag_structure(mpmodel.eval)
-  nnzj = length(jrows)
-  nnzh = length(hrows)
+  if :Jac in features
+    jrows, jcols = MathProgBase.jac_structure(mpmodel.eval)
+    nnzj = length(jrows)
+  else
+    jrows = Int[]
+    jcols = Int[]
+    nnzj = 0
+  end
+  if :Hess in features
+    hrows, hcols = MathProgBase.hesslag_structure(mpmodel.eval)
+    nnzh = length(hrows)
+  else
+    hrows = Int[]
+    hcols = Int[]
+    nnzh = 0
+  end
 
   meta = NLPModelMeta(nvar,
                       x0=mpmodel.x,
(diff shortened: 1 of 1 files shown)


dpo:
@vepiteski @abelsiqueira This is still incomplete but it seemed to work on a quick example:
```julia
julia> using OptimizationProblems
julia> prob = penalty2(5000);
julia> using NLPModels
julia> model = MathProgNLPModel(prob, features=[:Grad, :HessVec]);
julia> @time model = MathProgNLPModel(prob, features=[:Grad, :HessVec]);
  0.021754 seconds (517 allocations: 19.510 MB, 37.99% gc time)
```
We need:
- [ ] tests
- [ ] a check in `grad()`, `jac()` and `hess()` that `:Grad`, `:Jac` and `:Hess` were in the features requested (or that that capability is somehow available)

dpo:
How far do we want to go with this? It turns out that only `:Hess` has an effect. JuMP will generate the Jacobian structure whether you specify `:Jac` or not. Do we want

1. the user to input the features they want to have access to and perform hard checks that a feature is enabled before evaluating something (e.g., check that `:Jac` was requested before evaluating the Jacobian), or
2. simply be able to activate/deactivate the Hessian for JuMP models.

If we want 1, then we should do it for all other model types. I can see how this would be useful to make sure that a code is not using information they claim they are not using (e.g., a DFO code that issues calls to `grad()` or a factorization-free code that evaluates the Hessian).

abelsiqueira:
For option 1 we would have to use a consistent naming, instead of the MPB one. There is the issue of JuMP using jac for jprods, also. So disabling the feature Jac would be only symbolical if Jprod is still available.

If AMPL, or another possible nlp interface, has similar options, then option 1 is a good idea. Otherwise, it may take much more work than is worth.

Hard checks are good even if only for MPB.

dpo:
I think the point of option 1 would only be to police solvers and make sure they don't use information that they shouldn't use.

vepiteski:
There seems to be an issue between features specification and NLPtoMBP:

julia> using OptimizationProblems
julia> using NLPModels
julia> using Ipopt

julia> model = MathProgNLPModel(genrose(),features = [:Grad, :Hess]);
julia> ipoptmodel = NLPtoMPB(model, 
                             IpoptSolver(hessian_approximation="limited-memory",
                                         print_level = 0)
                                         );
ERROR: ArgumentError: throw: too many arguments (expected 1)
 in jac_coord(::NLPModels.MathProgNLPModel, ::Array{Float64,1}) at /home/local/USHERBROOKE/dusj1701/.julia/v0.5/NLPModels/src/mpb_model.jl:175
 in jac_structure(::NLPModels.NLPModelEvaluator{NLPModels.MathProgNLPModel}) at /home/local/USHERBROOKE/dusj1701/.julia/v0.5/NLPModels/src/nlp_to_mpb.jl:29
 in loadproblem!(::Ipopt.IpoptMathProgModel, ::Int64, ::Int64, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Symbol, ::NLPModels.NLPModelEvaluator{NLPModels.MathProgNLPModel}) at /home/local/USHERBROOKE/dusj1701/.julia/v0.5/Ipopt/src/IpoptSolverInterface.jl:48
 in NLPtoMPB(::NLPModels.MathProgNLPModel, ::Ipopt.IpoptSolver) at /home/local/USHERBROOKE/dusj1701/.julia/v0.5/NLPModels/src/nlp_to_mpb.jl:150

julia>

dpo:
Yes, besides the fact that there's a syntax error in my `throw`s, that's what a hard check does. Here you did not request `:Jac` (and I know why; the problem is unconstrained), but for some reason Ipopt is still trying to evaluate it. That's what I meant by policing algorithms. Is this really what we want?

Keep in mind that JuMP will generate the Jacobian structure *whether or not* you request `:Jac` (see the Google Group discussion with Miles). In this case, the Jacobian structure will be empty, but it will still be generated.

dpo:
ps: I think I fixed the syntax errors.

abelsiqueira:
After https://github.com/JuliaOpt/Ipopt.jl/pull/71, Ipopt should work.

If we don't police the solvers, these errors will depend on the underlying definition of the model.
So I think the hard check is a good addition. On the other hand, we would need to mantain some other codes, like Ipopt.

abelsiqueira:
New Ipopt version 0.2.6 includes the fix for this. The following code should work now:

```
julia> using OptimizationProblems
julia> using NLPModels
julia> using Ipopt

julia> model = MathProgNLPModel(genrose(),features = [:Grad, :Hess]);
julia> ipoptmodel = NLPtoMPB(model,
IpoptSolver(hessian_approximation="limited-memory",
print_level = 0)
);
julia> solveProblem(ipoptmodel.inner)
```

dpo:
Do we want to pursue this?

abelsiqueira:
I'm inclined to say no.

dpo:
I agree. It seems unrealistic.

PR #65, opened 2016-10-23
use evaluator that allows us to count evals

dpo:
This simple change allows us to retrieve evaluation counters from a `AbstractNLPModel` after calling a `MathProgBase` solver. It should have been this way since the beginning.


dpo:
```julia
julia> using OptimizationProblems
julia> using NLPModels
julia> using Ipopt
julia> model = MathProgNLPModel(hs6());
julia> ipoptmodel = NLPtoMPB(model, IpoptSolver());
julia> MathProgBase.optimize!(ipoptmodel)
# ... ipopt output ...
julia> model.counters
NLPModels.Counters(7,7,7,0,0,8,0,0,6,0,0)
julia> neval_obj(model)
7
```

abelsiqueira:
Thanks

PR #71, opened 2017-01-05
Update to JuMP 0.15

abelsiqueira:


dpo:
Same error here: https://ci.appveyor.com/project/dpo/nlpmodels-jl/build/1.0.230/job/0i8el3axproipru5#L165

With Julia 0.4, there's a conflict in the JuMP version required. Newer versions of JuMP need Julia 0.5. Should we drop Julia 0.4?

abelsiqueira:
I think so. 0.6 is around the corner.

PR #77, opened 2017-01-28
Update badges and documentation

abelsiqueira:
In addition, I'm not sure how the stable docs are supposed to be generated, so I generated them by hand.
The modifications on this PR are already on the stable version:
http://juliasmoothoptimizers.github.io/NLPModels.jl/stable/


dpo:
Looks awesome. Many thanks for all this!!

PR #85, opened 2017-04-13
ForwardDiff only returns dense derivatives

dpo:
Because Jacobians and Hessians returned by ForwardDiff are always dense arrays, using `sparse()` on them is not safe because the number of nonzeros and their location appears to change with the iterations. Until ForwardDiff returns sparse derivatives, I think we have no choice but store them as dense.


abelsiqueira:
Thanks

PR #86, opened 2017-04-14
only use multipliers for general constraints

dpo:
Certain solvers (i.e., KNITRO :roll_eyes:) supply m+n multipliers to evaluate the Hessian and Hessian-vector products, including multipliers associated to the bound constraints.


abelsiqueira:
Thanks

PR #91, opened 2017-07-11
Nonlinear Least Squares

abelsiqueira:


src/NLSModels.jl
```diff
@@ -0,0 +1,301 @@
...
+    return new(Counters(), 0, 0, 0, 0, 0, 0)
+  end
+end
+
+nls_counters(nls :: AbstractNLSModel) = nls.counters
+
+function sum_counters(c :: NLSCounters)
+  s = sum_counters(c.counters)
+  for field in fieldnames(NLSCounters)
+    if field == :counters; continue; end
```
dpo: This could be a bit simpler: `field == :counter && continue`.

abelsiqueira: Ah yes. I was using `if A; B end` instead of `A && B` for some time. But now that I look back, I can't see a good reason why.

dpo:
💯 Thanks! This is a lot of work! Will this basically deprecate NLSModels.jl?

abelsiqueira:
Yes, NLSModels.jl should be all here.

src/viability_model.jl
```diff
@@ -0,0 +1,93 @@
+export ViabilityModel
```
dpo: Where does the name Viability come from? Do you mean Feasibility?

abelsiqueira: Yes, sorry. Brazilianizing again.

src/feasibility_model.jl
```diff
@@ -0,0 +1,93 @@
+export FeasibilityModel
+
+# TODO: Extend to handle bounds
+"""
+A viability model is created from a NLPModel of the form
```
dpo: Here's another "viability" :-).

src/feasibility_model.jl
```diff
@@ -0,0 +1,93 @@
...
+
+    min f(x)
+    s.t c(x) = 0
+
+by defining the function F(x) = c(x). If the problem has
+bounds on the variables or more constraints, an error
+is thrown.
```
dpo: Normally, what we call a feasibility problem is a problem with a constant objective function. It can have equality and inequality constraints. Its sole purpose is to identify a feasible point. I wonder if we could give this one another name to avoid confusion. Maybe FeasibilityResidual or something like that?

abelsiqueira:
Updated. I used FeasibilityResidual as the new name.

src/NLSModels.jl
```diff
@@ -144,7 +147,7 @@ end
 """
     Jx = jac_op_residual(nls, x)
 
-Computes J(x), the Jacobian of the residual at x, on Linear Operator form.
+Computes J(x), the Jacobian of the residual at x, on linear operator form.
```
dpo: `in`

dpo:
Thank you! 💯

PR #93, opened 2017-09-17
Add constraints to autodiff nls

abelsiqueira:


src/NLSModels.jl
```diff
@@ -21,6 +21,7 @@ type NLSCounters
 end
 
 nls_counters(nls :: AbstractNLSModel) = nls.counters
+counters(nls :: AbstractNLSModel) = nls_counters(nls).counters
```
dpo: Do we want the counters to be aliased this way? Won't it make it difficult to compare, e.g., the number of objective evaluations when running a solver on a set of problems?

src/autodiff_nlsmodel.jl
```diff
@@ -11,22 +12,37 @@ type ADNLSModel <: AbstractNLSModel
...
                     lvar :: Vector = fill(-Inf, length(x0)),
-                    uvar :: Vector = fill(Inf, length(x0))
+                    uvar :: Vector = fill(Inf, length(x0)),
+                    y0 :: Vector = Float64[],
+                    c :: Function = (args...)->throw(NotImplementedError("cons")),
+                    lcon :: Vector = Float64[],
+                    ucon :: Vector = Float64[]
                    )
   nvar = length(x0)
+  ncon = maximum([length(lcon); length(ucon); length(y0)])
```
dpo: I think it should be an error if they were not of the same length. It's too confusing.

abelsiqueira: I'll forbid empty lcon and ucon then, and default y0 to zeros.

src/NLSModels.jl
abelsiqueira: It's this way because of models like `SlackModel`, which uses the counters of the submodel. In particular I'm using it to increment the counters. One alternative would be create something like
`increment_counter(c :: Symbol)`, and remove all `counters(nlp)` and `nls_counters(nls)`.

dpo: What I mean is: do `NLSModel` instances also have `neval_obj`, etc., counters?

src/autodiff_nlsmodel.jl
dpo: Ok.

src/NLSModels.jl
abelsiqueira: [Yes](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/abfb806e1d1af903f2b2bfe5bdcd2ccb810478f0/src/NLSModels.jl#L43). We probably need to test whether all this referencing is being done right.

dpo:
Where does this stand?

abelsiqueira:
I'm rebasing and fixing what was mentioned before.

src/autodiff_nlsmodel.jl
```diff
@@ -61,3 +73,101 @@ function hprod_residual!(nls :: ADNLSModel, x :: AbstractVector, i :: Int, v ::
   Hiv[:] = ForwardDiff.hessian(x->nls.F(x)[i], x) * v
   return Hiv
 end
+
+function cons(nls :: ADNLSModel, x :: Vector)
+  increment!(nls, :neval_cons)
+  return nls.c(x)
+end
```
dpo: If we asked the user to supply `cons!` instead of `cons`, would we loose efficiency in the AD computations?

src/autodiff_nlsmodel.jl
```diff
@@ -61,3 +73,101 @@ function hprod_residual!(nls :: ADNLSModel, x :: AbstractVector, i :: Int, v ::
...
+function cons!(nls :: ADNLSModel, x :: Vector, c :: Vector)
+  increment!(nls, :neval_cons)
+  c[1:nls.meta.ncon] = nls.c(x)
+  return c
+end
+
+function jac_coord(nls :: ADNLSModel, x :: Vector)
+  increment!(nls, :neval_jac)
+  J = ForwardDiff.jacobian(nls.c, x)
+  return findnz(J)
```
dpo: Because `J` is always dense, perhaps we can return something like `([j for i = 1 : n for j = 1 : m], [i for i = 1 : n for j = 1 : m], J[:])` (not tested)? I'm not sure if that would be more or less efficient.

dpo:
Are there any obvious tests we could add? Maybe derivative checks against finite differences?

src/autodiff_nlsmodel.jl
abelsiqueira: I don't think so. Do we change this is general, instead of only for NLS AD?

src/autodiff_nlsmodel.jl
abelsiqueira: J is dense, but may have zeros, and `findnz` will not return those.

abelsiqueira:
We can add more tests in NLS consistency, but I need to implement constraints in the other NLS models first - which I already started.

src/autodiff_nlsmodel.jl
dpo: Ok

src/autodiff_nlsmodel.jl
dpo: The only other places would be SimpleNLPModel and ADNLPModels, no?

abelsiqueira: Yes. It should be simple.

abelsiqueira:
NLS-consistency checks implemented on #102.

src/autodiff_nlsmodel.jl
abelsiqueira: I've checked if using only `c!(x, cx)` would be possible, but it doesn't appear to be because of the Hessian. We could have the user give both `c` and `c!`, but I'm not sure that's a good idea.

dpo: I don't think that would be a good idea. What's the issue with the Hessian?

abelsiqueira: I'm computing the Hessian of the i-th constraint as
```julia
ForwardDiff.hessian(x->nlp.c(x)[i], x)
```
and couldn't make it work using `c!`.

dpo: That's odd because `cons!(x, c)` is supposed to return `c`, so I would have expected `x -> nlp.cons!(x, c)[i]` to work. But if it doesn't, oh well.

abelsiqueira: It gives some conversion error with Dual

abelsiqueira:
Error is the same Homebrew problem. This should be ready to merge.

dpo:
Thank you.

PR #94, opened 2017-09-20
Update how counters work

abelsiqueira:
Create increment! function.
Erase counters(nlp) functions.

### DIFF
--- a/src/NLSModels.jl
+++ b/src/NLSModels.jl
@@ -1,4 +1,4 @@
-export AbstractNLSModel, nls_meta, NLSCounters, nls_counters, reset!,
+export AbstractNLSModel, nls_meta, NLSCounters, reset!,
        residual, residual!, jac_residual, jprod_residual, jprod_residual!,
        jtprod_residual, jtprod_residual!, jac_op_residual, jac_op_residual!,
        hess_residual, hprod_residual, hprod_residual!, hess_op_residual,
@@ -20,7 +20,13 @@ type NLSCounters
   end
 end
 
-nls_counters(nls :: AbstractNLSModel) = nls.counters
+function increment!(nls :: AbstractNLSModel, s :: Symbol)
+  if s in fieldnames(NLSCounters)
+    setfield!(nls.counters, s, getfield(nls.counters, s) + 1)
+  else
+    setfield!(nls.counters.counters, s, getfield(nls.counters.counters, s) + 1)
+  end
+end
 
 function sum_counters(c :: NLSCounters)
   s = sum_counters(c.counters)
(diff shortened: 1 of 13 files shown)


dpo:
What is the rationale for all these changes?

abelsiqueira:
Clarity. To avoid writing things like
```julia
nls_counters(nls).counters.neval_hprod += 1
```
and use only
```julia
increment!(nls, :neval_hprod)
```

abelsiqueira:
I realized i use `counters(nlp)` on [ExecutionStats](https://github.com/JuliaSmoothOptimizers/Optimize.jl/pull/35), although that isn't merged yet.
Any suggestions on how to proceed?

src/NLSModels.jl
```diff
@@ -20,7 +20,13 @@ type NLSCounters
...
 end
 
-nls_counters(nls :: AbstractNLSModel) = nls.counters
+function increment!(nls :: AbstractNLSModel, s :: Symbol)
+  if s in fieldnames(NLSCounters)
+    setfield!(nls.counters, s, getfield(nls.counters, s) + 1)
+  else
+    setfield!(nls.counters.counters, s, getfield(nls.counters.counters, s) + 1)
+  end
+end
```
dpo: Sorry for the long delay here. This PR looks good to me except the part above. Would we really ask directly increment a counter in the underlying model? I think `NLSModel` itself should do it. For example, when we evaluate the residual in a FeasibilityModel, the constraint evaluation counter of the underlying model should automatically be updated. What do you think?

abelsiqueira: Here, `increment!` is checking whether to update the counter from `NLSCounters` or from `Counters`, but they're both from the NLSModel - i.e. a NLSModel has its NLSCounters but also the default Counters as if it were a usual NLPModel. The FeasibilityModel is currently updating both counters, e.g.:
```julia
nlp = ADNLPModel(x->dot(x,x), zeros(2), c=x->[sum(x)-1], lcon=[0.0], ucon=[0.0])
nls = FeasibilityResidual(nlp)
obj(nls, zeros(2))
```
After running this code, nls obj counter increases by one and nlp cons counter increases by one, and the increment is automatic. Also, nls residual counter increases by one, because of the way obj works, and this increment is also automatic.

dpo: I see. Is this will in progress or finalized?

dpo: (will -> still)

abelsiqueira: It's finalized. I'm gonna try to find another to way to use Counters on ExecutionStats.

PR #98, opened 2017-11-15
Add precompile

abelsiqueira:


dpo:
I wonder if this conflicts with the fact that we have optional dependencies. Could it be the cause of https://travis-ci.org/JuliaSmoothOptimizers/Optimize.jl/jobs/306039663#L2617 ?

abelsiqueira:
I don't think so. Changing `import MathProgBase` on `nlp_to_mpb.jl` to `using MathProgBase` removed that error, but I haven't found a reason why. There are still errors on JuMP in 0.7, so I'm not sure if that's all.

PR #106, opened 2017-12-21
Remove .documenter.enc

abelsiqueira:
I had to create new keys in github and travis for the new way to deploy docs. It's working already and this file is not needed anymore.


dpo:
Thanks!

PR #109, opened 2018-03-20
Complete rework merging NLP and NLS

abelsiqueira:
I started working on a single NLP with both objectives to see how it would feel. The diff is huge, so it's probably better to discuss only the approach.

I considered this general NLP problem form
```
min f(x) = ∑ᵢσ_fᵢ × fᵢ(x) + σ_nls × ½‖F(x)‖² + σ_ls × ½‖Ax - b‖²
s.t cℓ ≤ c(x) ≤ cu,  ℓ ≤   x  ≤ u
```

I also considered removing the SimpleNLPModel and indicating how to create a simple model by hand, because the transition is very annoying.


abelsiqueira:
@dpo Suggestions on the best way to present this for review? Since it's a complete rework, I can't make small changes from the current `master`. I could try making a commit deleting everything and adding things incrementally. Would this work?

PR #115, opened 2018-04-25
Make SlackModel work with NLS Models

abelsiqueira:


src/slack_model.jl
```diff
@@ -71,7 +72,15 @@ function SlackModel(model :: AbstractNLPModel)
...
   )
 
-  return SlackModel(meta, model)
+  nls_meta = if isa(model, AbstractNLSModel)
+    NLSMeta(model.nls_meta.nequ,
+            model.meta.nvar + ns,
+            [model.meta.x0; zeros(ns)])
+  else
+    NLSMeta(0, 0)
+  end
```
dpo: Shouldn't we have two different constructors instead of an `if`?

dpo: In fact, I'm a bit confused here. Why does SlackModel derive from a more specific type than AbstractNLPModel?

abelsiqueira: Deriving from NLS, Slack is also an NLP, so it doesn't break anything that was working before. I did this to avoid creating a SlackNLSModel duplicating all the code.
But I see that this can be prone to error. Do you think I should create SlackNLSModel?

dpo: Is it really needed? Why can't we just define `residual`, `jprod_residual`, etc., for `SlackModel`s? If the underlying model doesn't support those methods, there will be an error.

abelsiqueira: Defining the NLS api while Slack was an NLP worked.

dpo:
While this works, I'm a bit torn about what implementation makes the most sense. Having both `meta` and `nls_meta` looks a bit messy but perhaps it makes sense if we consider problems with the structure you describe in #109. Perhaps NLPModels should already have a `nls_meta` to take LS terms into account?!

An alternative (that doesn't go the way of #109) could be something in the spirit of (untested)
```julia
mutable struct SlackModel{M,P} <: AbstractNLPModel where M <: AbstractNLPMeta, P <: AbstractNLPModel
  meta :: M
  model :: P
end
```

What do you think?

abelsiqueira:
The current `nls_meta` only adds `nequ` to the model, so all NLS models have both `meta` and `nls_meta`. A simple solution would be to add `nequ` to `meta` and remove `nls_meta`. This would be similar to #109, but `nequ` would be a dummy parameter in current NLP models.

Another solution would be to give `nls_meta` the missing parameters, `lvar, uvar, lcon`, etc. This would be more in accordance with your proposed structure, and more in accordance with different Metas for different models.

dpo:
I do see your point now that a slack model build from a NLSModel should also be an NLSModel. How can we achieve that without defining a new type?

abelsiqueira:
To have `SlackModel(nlp) -> NLP` and `SlackModel(nls) -> NLS`, without defining a new type, I don't think is possible.

Another option, instead of parametrisation, would be to create 
- `AbstractSlackModel`
- `NLPSlackModel <: AbstractSlackModel`
- `NLSSlackModel <: AbstractSlackModel`
- `function SlackModel(nlp :: AbstractNLPModel) ... return NLPSlackModel`
- `function SlackModel(nls :: AbstractNLSModel) ... return NLSSlackModel`

This allows creating extra `nls_meta` for `NLSSlack`, and hides from the user that there are different models. The NLPModels API can be implemented for `AbstractSlackModel` and the NLS part only for the `NLSSlack`.

If #109 or similar gets implemented, though, almost everything created here will be deleted later.

dpo:
I agree that that would be a good solution. Before we launch into it, let me have a closer look at #109 because it makes a lot of sense.

dpo:
Thanks! I'm starting to wonder if it wouldn't make more sense to use dispatch to define `SlackModel` for various subtypes of `AbstractNLPModel`. That way, `SlackModel(nls::AbstractNLSModel)` could return an `NLSModel`. I have the same situation with `SlackModel(qp::AbstractQuadraticModel)`, which I would like to return another `QuadraticModel`. What do you think?

abelsiqueira:
I like the idea, but I think we can't do that for every model - e.g. CUTEst. But we can have both. I'm trying to create an example.

abelsiqueira:
I like the idea, but I think we can't do that for every model - e.g. CUTEst. But we can have both. I'm trying to create an example.

dpo:
CUTEstModels have nothing special (they're not quadratic models, they're not nonlinear least-squares models, etc.), so only the default SlackModel constructor should apply to them.

abelsiqueira:
Updated according to our talk.

src/slack_model.jl
```diff
@@ -37,77 +37,111 @@ mutable struct SlackModel <: AbstractNLPModel
...
+
 import Base.show
 # TODO: improve this!
 # show(nlp :: SlackModel) = show(nlp.model)
 
 # retrieve counters from underlying model
 for counter in fieldnames(Counters)
   @eval begin
-    $counter(nlp :: SlackModel) = $counter(nlp.model)
+    $counter(nlp :: SlackModels) = $counter(nlp.model)
```
dpo: Why are we saying `SlackModels` instead of `SlackModel` now?

abelsiqueira: `SlackModels=Union{SlackModel,SlackNLSModel}`

src/slack_model.jl
```diff
@@ -231,3 +278,105 @@ function hprod!(nlp :: SlackModel, x :: AbstractVector, v :: AbstractVector,
...
+function jtprod_residual(nlp :: SlackNLSModel, x :: AbstractVector, v :: AbstractVector)
+  n = nlp.model.meta.nvar
+  ns = nlp.meta.nvar - n
+  return [jtprod_residual(nlp.model, x[1:n], v); zeros(ns)]
+end
+
+function jtprod_residual!(nlp :: SlackNLSModel, x :: AbstractVector, v :: AbstractVector, Jtv :: AbstractVector)
+  n = nlp.model.meta.nvar
+  ns = nlp.meta.nvar - n
+  @views jtprod_residual!(nlp.model, x[1:n], v, Jtv[1:n])
```
dpo: Sometimes you use `@views` and sometimes not. I'm not entirely sure when it is necessary.

src/slack_model.jl
dpo: Sorry, I missed it.

src/slack_model.jl
abelsiqueira: `@views` is the same as adding `@view` individually. `@view` doesn't work on assignments, so when there is only one I use `@view`. May have missed some places, let me check again.

dpo:
Thank you!

PR #119, opened 2018-05-09
Nonlinear Least Squares MathProgBase models

abelsiqueira:
Thanks, @mlubin and @odow, for the help.

The `MathProgNLSModel` receives a `JuMP` model with `NLexpression`s and a list of those expressions, creating two internal `JuMP` models, one with the expressions as constraints and the other as the equivalent minimization of `0.5*||F(x)||^2`. The residual, and its **sparse** Jacobian are obtained from the constraints of the first model. The second model provides the usual NLP functions, without having to use the residual - one exception is the Hessian-vector product function, due to this bug: https://github.com/JuliaOpt/JuMP.jl/issues/1204.


src/NLPModels.jl
```diff
@@ -344,7 +344,7 @@ conscale(::AbstractNLPModel, ::AbstractVector) =
   throw(NotImplementedError("conscale"))
 
 if Pkg.installed("MathProgBase") != nothing
-  include("mpb_model.jl")
```
dpo: What happened to `mpb_model`?

src/nlsmpb_model.jl
```diff
@@ -0,0 +1,253 @@
...
+       obj, grad, grad!,
+       cons, cons!, jac_coord, jac, jprod, jprod!, jtprod, jtprod!,
+       hess_coord, hess, hprod, hprod!,
+       residual!, jac_residual, jprod_residual!, jtprod_residual!,
+       jac_op_residual, hess_residual, hprod_residual!, cons, cons!, jac_coord,
+       jac, jprod, jprod!, jtprod, jtprod!, hess, hess_coord, hprod, hprod!
+
+mutable struct MathProgNLSModel <: AbstractNLSModel
+  meta :: NLPModelMeta
+  nls_meta :: NLSMeta
```
dpo: Are we moving towards merging both kinds of models into one then?

src/nlsmpb_model.jl
```diff
@@ -0,0 +1,253 @@
...
+function hprod!(nls :: MathProgNLSModel, x :: Array{Float64}, v :: Array{Float64},
+    hv :: Array{Float64};
+    obj_weight :: Float64=1.0, y :: Array{Float64}=zeros(nls.meta.ncon))
+  MathProgBase.eval_hesslag_prod(nls.cmodel.eval, hv, x, v, obj_weight, y)
+  return hv
+end
+=#
+
+function hprod!(nls :: MathProgNLSModel, x :: Array{Float64}, v :: Array{Float64},
+    hv :: Array{Float64};
```
dpo: So this version will be removed when 1204 is fixed?

dpo:
This looks excellent, @abelsiqueira ! Many thanks!

src/NLPModels.jl
abelsiqueira: It's being `include`d by `nlsmpb_model.jl`. It uses many MPB things in common.

src/nlsmpb_model.jl
abelsiqueira: Not yet, but this model will transition smoothly, it seems.

src/nlsmpb_model.jl
abelsiqueira: Yes. Notice that this version increments `residual` and `j*prod_*`, while using `cmodel` doesn't require that.

src/NLPModels.jl
dpo: I missed that. Thanks!

src/nlsmpb_model.jl
dpo: Ok

src/nlsmpb_model.jl
dpo: 👍

PR #120, opened 2018-05-30
Support of user-defined functions added in MathProgNLSModel

Xx-fractale-xX:


test/runtests.jl
```diff
@@ -40,6 +40,13 @@ end
 @assert isa(hess_op(model, [0.]), LinearOperator)
 @assert isa(jac_op(model, [0.]), LinearOperator)
 
+# MathProgNLSModel with user-defined functions
+model = mgh35()
+@assert model.meta.name == "mgh35"
+
+model = mgh07()
+@assert model.meta.name == "mgh07"
```
dpo: Could we also test that user functions are accessible (e.g., by evaluating the residuals)?

test/mgh07.jl
```diff
@@ -0,0 +1,32 @@
...
+function mgh07()
+
+  nls  = Model()
+  x0   = [-1.0,  0.0,  0.0]
+  lvar = [-Inf, -Inf, -Inf]
+  uvar = [ Inf,  Inf,  Inf]
+  @variable(nls, lvar[i] <= x[i=1:3] <= uvar[i], start = x0[i])
+
+  function θ_aux(t)
+  	return (t > 0 ? 0.0 : 0.5)
```
dpo: This could we a one liner: `θ_aux(t) = t > 0 ? 0.0 : 0.5`.

test/mgh07.jl
```diff
@@ -0,0 +1,32 @@
...
+export mgh07
+
+"Helical valley function"
+function mgh07()
+
+  nls  = Model()
+  x0   = [-1.0,  0.0,  0.0]
+  lvar = [-Inf, -Inf, -Inf]
+  uvar = [ Inf,  Inf,  Inf]
+  @variable(nls, lvar[i] <= x[i=1:3] <= uvar[i], start = x0[i])
```
dpo: Is it necessary to give the bounds if there aren't any?

test/mgh35.jl
```diff
@@ -0,0 +1,47 @@
...
+  if m < n
+    warn(": number of function must be ≥ number of variables. Adjusting to m = n")
+    m = n
+  end
+
+  nls  = Model()
+  x0   = collect(1:n)/(n+1)
+  lvar = -Inf*ones(n)
+  uvar =  Inf*ones(n)
+  @variable(nls, lvar[i] <= x[i=1:n] <= uvar[i], start = x0[i])
```
dpo: The bounds don't seem necessary here.

test/mgh35.jl
```diff
@@ -0,0 +1,47 @@
...
+    else
+      return 2*x*Tsim(x,n-1) - Tsim(x,n-2)
+    end
+  end
+
+  Ts = Vector{Function}(n)
+  Tnames = Vector{Symbol}(n)
+  for i = 1:n
+    Ts[i] = x -> Tsim(2*x-1, i)
+    Tnames[i] =  Symbol("Ts",i)
```
dpo: `gensym()` could be useful here to make sure the symbol doesn't conflict with another.

Xx-fractale-xX: I thought that symbols were only accessible by the model now. But you're right, it's preferable to use ```gensym()``` to be sure.

test/mgh07.jl
Xx-fractale-xX: No, I will delete lvar and uvar for the two models.

Xx-fractale-xX:
I create tests of  ```MathProgNLSModels```  with small modification of ```runtests.jl``` from ```NLSProblems``` repository in ```test_mathprognlsmodel.jl```. I add also two models ```hs30.jl``` and ```hs43.jl```, because ```mgh07.jl``` and ```mgh35.jl``` don't have constraints. Tests will be most exhaustive with them.

test/hs30.jl
```diff
@@ -0,0 +1,24 @@
...
+
+export hs30
+
+"Hock-Schittkowski problem 30 in NLS format"
+function hs30()
+
+  model = Model()
+  lvar = [1.0; -10.0; -10.0]
+  @variable(model, lvar[i] <= x[i=1:3] <= 10, start=1.0)
+  @NLexpression(model, F[i=1:3], x[i] + 0.0)
```
dpo: Why do we need to add `+ 0.0` here?

test/mgh07.jl
```diff
@@ -0,0 +1,27 @@
...
+  nls  = Model()
+  x0   = [-1.0,  0.0,  0.0]
+  @variable(nls, x[i=1:3], start = x0[i])
+
+  θ_aux(t) = (t > 0 ? 0.0 : 0.5)
+  JuMP.register(nls, :θ_aux, 1, θ_aux, autodiff=true)
+
+  @NLexpression(nls, F1, 10*(x[3] - 10*(atan(x[2]/x[1])/(2*π) + θ_aux(x[1]))))
+  @NLexpression(nls, F2, 10*(sqrt(x[1]^2 + x[2]^2) - 1.0))
+  @NLexpression(nls, F3, 1*x[3])
```
dpo: And here, do we need to multiply by 1?

test/hs30.jl
abelsiqueira: `x[i]` alone raises an error. It's treated differently by JuMP, but I did not investigate if it can be solved on our side.
`1 * x[i]` below also works, apparently.

abelsiqueira:
I think is an approximation issue. `gs` is an array of gradients, `gs[i]` is the gradient computed by `grad` and `g` is computed by `objgrad`, for the same model. They should be equal, without the `all()`. Maybe on 32bit some model has a slight deviation in the computation.

dpo:
It sounds like is would be safer to use `isapprox` than `==` then?!

abelsiqueira:
Yes, we could change all `==` to `isapprox`

test/mgh07.jl
Xx-fractale-xX: In the file ```jump_model.jl```, the function ```MathProgNLSModel``` only accept ```Vector{JuMP.NonlinearExpression}```.

``` julia
function MathProgNLSModel(cmodel :: JuMP.Model,
                          F :: Vector{JuMP.NonlinearExpression}; kwargs...)
  # Hessian-Vector products don't work:
  # https://github.com/JuliaOpt/JuMP.jl/issues/1204
  @NLobjective(cmodel, Min, 0.5 * sum(Fi^2 for Fi in F))
  JuMP.setsolver(cmodel, ModelReader())
  cmodel.internalModelLoaded || JuMP.build(cmodel)
  ev = cmodel.internalModel.eval

  Fmodel = JuMP.Model()
  @NLobjective(Fmodel, Min, 0.0)
  Fmodel.nlpdata.user_operators = cmodel.nlpdata.user_operators
  @variable(Fmodel, x[1:MathProgBase.numvar(cmodel)])
  for Fi in F
    expr = ev.subexpressions_as_julia_expressions[Fi.index]
    replace!(expr, x)
    expr = :($expr == 0)
    JuMP.addNLconstraint(Fmodel, expr)
  end
  JuMP.setsolver(Fmodel, ModelReader())
  Fmodel.internalModelLoaded || JuMP.build(Fmodel)

  return MathProgNLSModel(Fmodel.internalModel,
                          cmodel.internalModel; kwargs...)
end
```
And JuMP want that we use ```@expression``` instead of ```@NLexpression``` when we have only a JuMP variable as expression. It's why we use the trick ```1*x[i]``` or ```x[i] + 0.0``` to allow it.

A solution to avoid it could be to accept both expression like that:
``` julia 
function MathProgNLSModel(cmodel :: JuMP.Model,
                          F :: Vector{JuMP.NonlinearExpression,JuMP.AffExpr}; kwargs...)
  # Hessian-Vector products don't work:
  # https://github.com/JuliaOpt/JuMP.jl/issues/1204
  @NLobjective(cmodel, Min, 0.5 * sum(Fi^2 for Fi in F))
  JuMP.setsolver(cmodel, ModelReader())
  cmodel.internalModelLoaded || JuMP.build(cmodel)
  ev = cmodel.internalModel.eval

  Fmodel = JuMP.Model()
  @NLobjective(Fmodel, Min, 0.0)
  Fmodel.nlpdata.user_operators = cmodel.nlpdata.user_operators
  @variable(Fmodel, x[1:MathProgBase.numvar(cmodel)])
  for Fi in F
  	if typeof(Fi) == JuMP.NonlinearExpression
    	expr = ev.subexpressions_as_julia_expressions[Fi.index]
    	replace!(expr, x)
    	expr = :($expr == 0)
    	JuMP.addNLconstraint(Fmodel, expr)
    else
    	c = GenericRangeConstraint(Fi,0.0,0.0)
    	JuMP.addconstraint(Fmodel, c)
    end
  end
  JuMP.setsolver(Fmodel, ModelReader())
  Fmodel.internalModelLoaded || JuMP.build(Fmodel)

  return MathProgNLSModel(Fmodel.internalModel,
                          cmodel.internalModel; kwargs...)
end
```
But I don't know the syntax to declare a vector that accept both types ```Vector{JuMP.NonlinearExpression,JuMP.AffExpr}```.

abelsiqueira: I think something with [parametric types](https://docs.julialang.org/en/stable/manual/types/#Parametric-Types-1).

Xx-fractale-xX:
I forgot to modify ```consistency.jl```, but before I change ```==``` in it, I used ```≈``` , equivalent to isapprox(x,y;rtol=sqrt(eps), atol=0). But maybe it's better to use isapprox with ``` atol != 0 ```?

dpo:
Ideally, I would prefer if that were in a different pull request, as it is separate from the topic of this one. I think @abelsiqueira used `isapprox` with a nonzero `atol`, so we should be consistent and use the same everywhere. However, the tolerance should take the arithmetic into account (32 bit vs. 64).

dpo:
(oops, I didn't mean to close this)

dpo:
The tests pass but this branch needs to be rebased against master.

dpo:
Thank you!!

PR #121, opened 2018-06-05
Improvement of tests

Xx-fractale-xX:


test/consistency.jl
```diff
@@ -63,20 +63,20 @@ function consistent_functions(nlps; nloops=100, rtol=1.0e-8, exclude=[])
       gmin = minimum(map(norm, gs))
       for i = 1:N
         for j = i+1:N
-          @test isapprox(gs[i], gs[j], atol=rtol * max(gmin, 1.0))
+          @test isapprox(gs[i], gs[j], atol = rtol * max(gmin, 1.0))
```
dpo: Please try to keep your pull requests focused on a single issue. This kind of update makes the change set appear much larger than it really is. Usually, we don't use a space when passing arguments to a function (although it's the recommended Julia style).

Xx-fractale-xX: It wasn't voluntary to keep commits of the other pull request. I created a new branch on the fork repository that also take into account last commits.
I deleted useless spaces like you recommended. It will be easier to review now and the update concern only that need to be modify.
I also squashed all the commits to have a clean history.

dpo:
Thanks very much! I'm not sure what the problem is on macOS, but it's not related to this pull request.

PR #124, opened 2018-07-30
Modification of grad! function in ADNLPModels

### LINKED ISSUE #123: grad! returns wrong type in ADNLPModels
g = grad!(nlp, x, g) returns a SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}
but  g is an    Array{Float64,1}

causing the next call to grad! issue the error grad! not implemented

JPD

Xx-fractale-xX:
fixes #123


abelsiqueira:
Thank you

PR #129, opened 2018-09-06
Upper bound Julia to before 1.0

abelsiqueira:
For https://github.com/JuliaLang/METADATA.jl/pull/17800#issuecomment-419236990


dpo:
I see the point of upper bounding Julia in `REQUIRE` but what's wrong with testing on Travis and Appveyor?

abelsiqueira:
Travis and appveyor restored.

dpo:
Thank you!

PR #132, opened 2018-09-07
Change Vector to AbstractVector and add test for @view

### LINKED ISSUE #131: Can't evaluate problem functions with views
```julia
julia> nlp = ADNLPModel(x -> dot(x, x), zeros(2))
julia> x = ones(3);
julia> obj(nlp, @view x[1:2])
ERROR: MethodError: no method matching obj(::NLPModels.ADNLPModel, ::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true})
```

This might also apply to other problem types.

abelsiqueira:
Closes #131


src/autodiff_nlsmodel.jl
```diff
@@ -152,19 +152,19 @@ function hess(nls :: ADNLSModel, x :: Vector; obj_weight = 1.0, y :: Vector = []
...
-function hess_coord(nls :: ADNLSModel, x :: Vector; obj_weight = 1.0, y :: Vector = [])
+function hess_coord(nls :: ADNLSModel, x :: AbstractVector; obj_weight = 1.0, y :: AbstractVector = [])
   H = hess(nls, x, obj_weight=obj_weight, y=y)
   return findnz(H)
 end
 
-function hprod(nls :: ADNLSModel, x :: Vector, v :: Vector;
-    obj_weight = 1.0, y :: Vector = [])
+function hprod(nls :: ADNLSModel, x :: AbstractVector, v :: AbstractVector;
+    obj_weight = 1.0, y :: AbstractVector = [])
```
dpo: Should the default value be `Float64[]`?

src/autodiff_nlsmodel.jl
```diff
@@ -152,19 +152,19 @@ function hess(nls :: ADNLSModel, x :: Vector; obj_weight = 1.0, y :: Vector = []
...
   Hv = zeros(nls.meta.nvar)
   return hprod!(nls, x, v, Hv, obj_weight=obj_weight, y=y)
 end
 
-function hprod!(nls :: ADNLSModel, x :: Vector, v :: Vector, Hv :: Vector;
-    obj_weight = 1.0, y :: Vector = [])
+function hprod!(nls :: ADNLSModel, x :: AbstractVector, v :: AbstractVector, Hv :: AbstractVector;
+    obj_weight = 1.0, y :: AbstractVector = [])
```
dpo: Same here?

src/lls_model.jl
```diff
@@ -82,53 +82,53 @@ function hprod_residual!(nls :: LLSModel, x :: AbstractVector, i :: Int, v :: Ab
...
 
-function jtprod!(nls :: LLSModel, x :: Vector, v :: Vector, Jtv :: Vector)
+function jtprod!(nls :: LLSModel, x :: AbstractVector, v :: AbstractVector, Jtv :: AbstractVector)
   increment!(nls, :neval_jtprod)
   Jtv[1:nls.meta.nvar] = nls.C' * v
   return Jtv
 end
 
-function hess(nls :: LLSModel, x :: Vector; obj_weight = 1.0, y :: Vector = [])
+function hess(nls :: LLSModel, x :: AbstractVector; obj_weight = 1.0, y :: AbstractVector = [])
```
dpo: Same here, an several other places.

dpo:
Many thanks for this!!! I notice that sometimes, we use `AbstractVector` and sometimes `AbstractVector{Float64}`. Is there a reason?

abelsiqueira:
I don't remember if we discussed something about it. I noticed that `NLPModelMeta` uses `Array{Float64,1}`, and I think we used `Vector{Float64}` at first because of it, but eventually added functions with `Vector`.

I think changing `NLPModelMeta` to `AbstractVector` is in order?

dpo:
I think it's inefficient to have abstract types inside a struct. But I think we should use either AbstractVector or AbstractVector{Float64} everywhere.

abelsiqueira:
To handle different precision, we can't use `Float64` and to handle views, we need abstract. So I think it needs to be `AbstractVector`. Unless we don't allow views inside `NLPModelMeta`.
PS. Why is it ineficient?

dpo:
Ok for `AbstractVector`. Thanks!

https://docs.julialang.org/en/v0.7.0/manual/performance-tips/#Avoid-containers-with-abstract-type-parameters-1

dpo:
Another thought. Eventually, we'll want to parameterize `NLPModelMeta` and `NLPModel` to indicate what precision is used. I suggest we leave `Vector` in `NLPModelMeta` for now. Those don't need to be views or subarrays.

abelsiqueira:
Done.

dpo:
Thank you! Let's go with the last version of NLPModels compatible with 0.6.

PR #133, opened 2018-09-21
phase out Julia 0.6

dpo:
suppress many (not all) Julia 0.7 warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,12 @@ os:
   - osx
 
 julia:
-  - 0.6
   - 0.7
   - 1.0
   - nightly
 
 matrix:
   allow_failures:
-    - julia: 1.0
     - julia: nightly
 
 sudo: required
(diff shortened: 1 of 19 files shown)


dpo:
Not so sure where the remaining warnings come from.

dpo:
Julia 1.0 says
```
ERROR: LoadError: too many parameters for type
```
Not sure I understand what the error is.
https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/431711229#L276

abelsiqueira:
I think it's related to the new LinearOperators. I'll try to update this now. May need to update FastClosures too.

dpo:
NLPModels will also need to depend on FastClosures to create `hess_op` and `jac_op`.

abelsiqueira:
Rebased #134 here.

dpo:
It even passes on nightly!

PR #134, opened 2018-09-22
Other Julia 0.7/1.0 updates

abelsiqueira:
Additional changes to #133.

This PR targets the branch `julia-0.7`, so there is one additional step to see the build.

Let me know if you prefer something different, like a new PR superseeding #133, or a commit directly on branch `julia-0.7`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.7 2.0
-LinearOperators 0.4.0
-ForwardDiff 0.6.0
+LinearOperators 0.5.0
+ForwardDiff 0.9.0
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,7 +34,8 @@ install:
 build_script:
 # Need to convert from shallow to complete for Pkg.clone to work
   - IF EXIST .git\shallow (git fetch --unshallow)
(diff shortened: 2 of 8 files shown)


abelsiqueira:
The build for this branch is here: https://travis-ci.org/abelsiqueira/NLPModels.jl/builds/431773428
I clone a personal version of FastClosures.

We need https://github.com/c42f/FastClosures.jl/pull/8 and a release so that FastClosures count as working on 1.0. This is due to https://discourse.julialang.org/t/package-compatibility-caps/15301, effectively prohibiting the use of FastClosures on 1.0.

dpo:
Thanks @abelsiqueira ! I'll update #133 when FastClosures is updated.

Regarding the changes to NLSModel, lls_model, etc., I hesitated because we're not sure whether the Hessian is sparse or dense, are we? If it's sparse, your current changes will fail.

abelsiqueira:
I've fixed the changes on `_coord` to address that. On `ADNLPModel` the matrices are dense, otherwise I've separated in cases.

On #133, you mentioned `FastClosures` on `hess_op` and `jac_op`, is 3b3deb7 what you mean?

dpo:
@abelsiqueira Do you want to push your extra changes to #133?

PR #136, opened 2018-10-03
require FastClosures

dpo:


abelsiqueira:
Thanks. Do we create a new release now or wait a bit?

dpo:
Well, technically, the current release is missing a dependency. It will always be there because of LinearOperators, so I don't think it's crucial to release another right away.

PR #137, opened 2018-11-13
Update docs to new Documenter

abelsiqueira:


dpo:
Many thanks for all this! It looks good to me, though I don't know if I can see the docs generated somewhere before merging?!

abelsiqueira:
Not by default, but I'll generate a manual version.

abelsiqueira:
Here: http://abelsiqueira.github.io/NLPModels.jl/tmp/

dpo:
Thank you!

PR #139, opened 2018-11-23
Slack remodel

abelsiqueira:
Built on top of #115, please review only the last commit.

This adds `SlackModel(::ADNLPModel)` which returns `ADNLPModel`.

### DIFF
--- a/src/feasibility_residual.jl
+++ b/src/feasibility_residual.jl
@@ -5,11 +5,16 @@ export FeasibilityResidual
 A feasibility residual model is created from a NLPModel of the form
 
     min f(x)
-    s.t c(x) = 0
+    s.t cℓ ≤ c(x) ≤ cu
+        bℓ ≤   x  ≤ bu
 
-by defining the function F(x) = c(x). If the problem has
-bounds on the variables or more constraints, an error
-is thrown.
+by creating slack variables s and defining F(x,s) = c(x) - s. The resulting NLS problem is
+
+    min ¹/₂‖c(x) - s‖²
+        bℓ ≤ x ≤ bu
+        cℓ ≤ s ≤ bu
+
+This is done using SlackModel first, and then defining the NLS.
 """
 mutable struct FeasibilityResidual <: AbstractNLSModel
   meta :: NLPModelMeta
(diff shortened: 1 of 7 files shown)


src/slack_model.jl
```diff
@@ -68,16 +68,23 @@ function slack_meta(meta :: NLPModelMeta)
...
-function SlackModel(model :: AbstractNLPModel)
+"""
+    ConvertToSlackModel(model)
+
+Convert an NLPModel or NLSModel to a SlackModel, without using specialized information
+about the type of `model`.
+
+You may want `SlackModel(model)` instead.
+"""
+function ConvertToSlackModel(model :: AbstractNLPModel)
```
dpo: Why would we ever want this?

src/slack_model.jl
```diff
@@ -372,3 +392,26 @@ function hess_op_residual!(nls :: SlackModel, x :: AbstractVector, i :: Int, Hiv
...
+  c(x) = begin
+    cx = nlp.c(@view x[1:n])
+    cx[nlp.meta.jlow] .-= @view x[n+1:n+nlow]
+    cx[nlp.meta.jupp] .-= @view x[n+nlow+1:n+nlow+nupp]
+    cx[nlp.meta.jrng] .-= @view x[n+nlow+nupp+1:n+nlow+nupp+nrng]
+    return cx
+  end
+
+  return ADNLPModel(meta, deepcopy(nlp.counters), f, c)
+end
```
dpo: Yes, that's the idea! I think this is the way to go.

src/slack_model.jl
abelsiqueira: I'm not sure we will but currently, it's used for testing on `test/test_slack_model.jl`. Possibly won't need it after this is finished.

src/feasibility_residual.jl
```diff
@@ -5,11 +5,16 @@ export FeasibilityResidual
...
-    s.t c(x) = 0
+    s.t cℓ ≤ c(x) ≤ cu
+        bℓ ≤   x  ≤ bu
 
-by defining the function F(x) = c(x). If the problem has
-bounds on the variables or more constraints, an error
-is thrown.
+by creating slack variables s and defining F(x,s) = c(x) - s. The resulting NLS problem is
+
+    min ¹/₂‖c(x) - s‖²
```
dpo: It seems f(x) disappeared?!

abelsiqueira: Yes, `f` is ignored on for the NLS model. I could change to `min 0`.

abelsiqueira:
Dropping in favor of #115

PR #143, opened 2019-01-13
LLSModel constraints are linear

### LINKED ISSUE #126: lls_model is NOT linear
When querying a lls_model, linearly_constrained(lls_model) returns false, and on my small example, ncon=8, nlin=0.

abelsiqueira:
Closes #126


dpo:
Great!

PR #146, opened 2019-01-17
Implement NLStoCons

abelsiqueira:
Need to rebase after #115


abelsiqueira:
This is ready for review

abelsiqueira:
This implements `NLStoCons` which converts an `NLSModel` by setting `F(x) = r` and using the sum of r squares as residual. The model is itself an `NLSModel`.

dpo:
This looks great, but I wonder if we can find a better name than NLStoCons?! I've heard people referring to this transformation as "dualizing" the least-squares problem.

abelsiqueira:
I don't like the name also (though I started with `NLStoNLPwCons`), but I didn't understand the "dualizing".

dpo:
I think "dualizing" simply refers to moving a term to the constraints. But it's not a great name either. How about `FeasibilityNLS`? Or `FeasibilityFormNLS`?

abelsiqueira:
I like `FeasibilityFormNLS`.

src/NLPModels.jl
```diff
@@ -365,6 +365,7 @@ include("autodiff_model.jl")
 include("simple_model.jl")
 include("slack_model.jl")
 include("qn_model.jl")
+include("nls_to_cons.jl")
```
dpo: Would you mind renaming this file? Thanks !

test/runtests.jl
```diff
@@ -78,3 +78,4 @@ test_view_subarrays()
 include("test_nlsmodels.jl")
 include("nls_consistency.jl")
 include("test_generic_coord.jl")
+include("test_nls_to_cons.jl")
```
dpo: And this one

dpo:
Great, thank you! Go ahead and merge after the file renames.

PR #147, opened 2019-01-17
FeasibilityResidual of models with inequalities

abelsiqueira:
Need to rebase after #115


abelsiqueira:
This is ready now.

abelsiqueira:
This allows building `FeasibilityResidual` models from problems with inequality, by using `SlackModel` first and then applying `FeasibilityResidual`.

PR #145, opened 2019-01-17
Small fix and test

abelsiqueira:
Small fix on `hess_coord` of general `NLSModels`.
Add more tests to `LLSModel`.

PR #148, opened 2019-01-17
 Improve testing for view subarray argument and related fixes

abelsiqueira:
Needs to be rebased after #115 ~~and #145~~.

### DIFF
--- a/test/test_view_subarray.jl
+++ b/test/test_view_subarray.jl
@@ -1,19 +1,217 @@
-function test_view_subarray()
-  snlp = SimpleNLPModel(x -> sum(x.^4), ones(3),
-                        g=x -> 4x.^3,
-                        c=x -> [dot(x, x) - 4.0; sum(x) - 1.0],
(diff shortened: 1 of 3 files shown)


test/test_view_subarray.jl
```diff
@@ -1,19 +1,217 @@
...
+        @test rows1 == rows2
+        @test cols1 == cols2
+        @test vals1 ≈ vals2
+      end
+
+      # Inplace methods can have input and output as view, so 4 possibilities
+      for J = Vidxs
+        gv = @view g2[J]
+        grad!(nlp, x[I],  g)
+        grad!(nlp, x[I], gv); @test g ≈ g2[J]
```
dpo: Should this compare g to gv instead of g2?

abelsiqueira: `gv` is the view for `g2[J]`, so they should be the same. Do you want me to change?

dpo: Ah, sorry. No need to change. It looks good.

PR #150, opened 2019-01-18
Improve LLS consistency checks

abelsiqueira:
These are some additional tests on the consistency of `LLSModel`:
- `SlackNLSModel` of `LLSModel` vs `SlackNLSModel` of `ADNLSModel`;
- `LLSModel` with matrix vs `LLSModel` with `LinearOperator`
Some large changes in `nls_consistency` are actually just additional `if` checking if some tests are to be excluded.

### DIFF
--- a/test/nls_consistency.jl
+++ b/test/nls_consistency.jl
@@ -1,3 +1,5 @@
+import LinearAlgebra: I
+
 include("consistency.jl")
 
 function consistent_nls_counters(nlss)
@@ -12,7 +14,7 @@ function consistent_nls_counters(nlss)
   @test all(V .== V[1])
 end
 
-function consistent_nls_functions(nlss; nloops=10, rtol=1.0e-8)
+function consistent_nls_functions(nlss; nloops=10, rtol=1.0e-8, exclude=[])
   N = length(nlss)
   n = nls_meta(nlss[1]).nvar
   m = nls_meta(nlss[1]).nequ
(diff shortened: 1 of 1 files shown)


dpo:
I'm really not a fan of random tests. They are probably the cause of the failure in #147.

abelsiqueira:
Change the tests introduced here. Should I open an issue to change all other `rand` tests?

dpo:
I thin so, because they fail randomly...

test/nls_consistency.jl
```diff
@@ -23,77 +26,85 @@ function consistent_nls_functions(nlss; nloops=10, rtol=1.0e-8)
...
-    J_ops = Any[jac_op_residual(nls, x) for nls in nlss]
-    Jv, Jtv = zeros(m), zeros(n)
-    J_ops_inplace = Any[jac_op_residual!(nls, x, Jv, Jtv) for nls in nlss]
+    if intersect([jac_op_residual, jprod_residual, jtprod_residual],  exclude) == []
+      J_ops = Any[jac_op_residual(nls, x) for nls in nlss]
+      Jv, Jtv = zeros(m), zeros(n)
+      J_ops_inplace = Any[jac_op_residual!(nls, x, Jv, Jtv) for nls in nlss]
 
-    v = rand(n)
+      v = rand(n)
```
dpo: There's still a random vector here, and another one below. I would replace them with [1, -1, 1, -1, ...].

vepiteski: I would prefer some random instance(s) for example using one or several seeds to have reproducible results. A simple test with say [1, -1, ...] may fail to exhibit issues because of its evident symmetry and specificity.

abelsiqueira: Sorry, missed those because of `git diff -w`.

dpo: I don't think a seed will guarantee reproducible results on various platforms. The vector [1, -1, 1, -1, ...] is often used in place of a random vector and will give reproducible results. Yes it may fail to exhibit some issues, but it will not give false positives.

test/nls_consistency.jl
```diff
@@ -23,77 +26,85 @@ function consistent_nls_functions(nlss; nloops=10, rtol=1.0e-8)
   for k = 1:nloops
     x = 10 .* (rand(n) .- 0.5)
```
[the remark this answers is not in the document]
abelsiqueira: There are actually quite a few other places with randomness yet, but they left the scope of this PR. I'll do a PR removing the randomness first, then rebase this one.

PR #153, opened 2019-01-21
Create new hess_residual which computes sum of Hessians

abelsiqueira:
New `hess_residual(nlp, x, v)` computes `\sum_{i=1}^{n_e} v_i \nabla^2 F_i(x)`, and increments `:neval_hess_residual`.
Old `hess_residual(nlp, x, i)` computes `\nabla^2 F_i(x)`, and now increments the new `:neval_ihess_residual`.

### DIFF
--- a/src/NLSModels.jl
+++ b/src/NLSModels.jl
@@ -1,7 +1,7 @@
 export AbstractNLSModel, nls_meta, NLSCounters, reset!,
        residual, residual!, jac_residual, jprod_residual, jprod_residual!,
        jtprod_residual, jtprod_residual!, jac_op_residual, jac_op_residual!,
-       hess_residual, hprod_residual, hprod_residual!, hess_op_residual,
+       hess_residual, jth_hess_residual, hprod_residual, hprod_residual!, hess_op_residual,
        hess_op_residual!, NotImplementedError
 
 abstract type AbstractNLSModel <: AbstractNLPModel end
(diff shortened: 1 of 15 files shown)


src/NLSModels.jl
```diff
@@ -183,14 +184,24 @@ function jac_op_residual!(nls :: AbstractNLSModel, x :: AbstractVector,
...
+Computes the linear combinations of the Hessians of the residuals at x with coefficients
+`v`.
 """
-function hess_residual(nls :: AbstractNLSModel, x :: AbstractVector, i :: Int)
+function hess_residual(nls :: AbstractNLSModel, x :: AbstractVector, v :: AbstractVector)
   throw(NotImplementedError("hess_residual"))
 end
 
+"""
+    Hi = jth_hess_residual(nls, x, i)
```
dpo: Maybe use `j` instead of `i`?

abelsiqueira: I've noticed that I've used `i` in a lot of places for Hessian related things, though `jth` existed for other things. Do you think it's better to update all to `j`?

dpo: It's a detail but it may be clearer in the docstring (in the functions, it doesn't really matter).

abelsiqueira: Updated.

dpo:
Many thanks for all this!!!

PR #156, opened 2019-01-22
Implement jac_coord_residual and hess_coord_residual

abelsiqueira:
This is built on top of #153, needs rebasing after that.

This implements coordinate information for the Jacobian and Hessian of the residual function: number of nonzeros and coordinate function call, and updates the existing coordinate function calls for consistency.
- coord functions must be implemented manually now for each model, instead of having a generic fallback.
- All NLP models must implement jac_coord and hess_coord, always returning the same structure and providing `nnzj` and `nnzh` correctly.
- All NLS models must implement jac_coord_residual and hess_coord_residual, always returning the same structure and providing `nls_meta.nnzj` and `nls_meta.nnzh` correctly.
- The exception to the rules above are the `Simple` models.
- The conditions above imply that sometimes the coordinate format will store zeroes.
- Some models, like `ADNLPModel`, `ADNLSModel` and `LLSModel` with a dense matrix, always return "dense" coordinates, i.e., all elements are returns, even if they are null.
- By default NLS models can be viewed as NLPs with objective of the form `f(x) = |F(x)|^2`. This means the Hessian not trivially available for every model, so **`hess_coord` is not implemented** for these methods.


src/feasibility_residual.jl
```diff
@@ -46,12 +46,14 @@ end
 
 function residual(nls :: FeasibilityResidual, x :: AbstractVector)
   increment!(nls, :neval_residual)
-  return cons(nls.nlp, x)
+  return cons(nls.nlp, x) - nls.nlp.meta.lcon
 end
```
abelsiqueira: @dpo did we decide that equalities must have `lcon = 0`? I don't think so, right?

abelsiqueira:
A `FeasibilityForm` of a `FeasibilityResidual` can get better information, so I added another commit.

src/feasibility_residual.jl
dpo: No, not necessarily.

src/autodiff_model.jl
```diff
@@ -114,6 +114,13 @@ function jac(nlp :: ADNLPModel, x :: AbstractVector)
   return ForwardDiff.jacobian(nlp.c, x)
 end
 
+function jac_coord(nlp :: ADNLPModel, x :: AbstractVector)
+  Jx = jac(nlp, x)
+  m, n = nlp.meta.ncon, nlp.meta.nvar
+  I = ((i,j) for i = 1:m, j = 1:n)
+  return (getindex.(I, 1)[:], getindex.(I, 2)[:], Jx[:])
```
abelsiqueira: Julia 0.7 is failing because `I` is a generator instead of an array. It's a small cost, but it could be reason to drop 0.7. Since we already made the transition from 0.6, it shouldn't affect anything.

dpo: I agree but I still find 0.7 useful in the sense that it gives warnings and tells you how to change your code to make it compatible with 1.0. Could we just add an `if VERSION ...` for the moment?

PR #157, opened 2019-01-24
Update ADNLPModel and ADNLSModel to work with multiple precision

abelsiqueira:


src/NLSModels.jl
```diff
@@ -93,8 +93,8 @@ include("feasibility_residual.jl")
 
 Computes F(x), the residual at x.
 """
-function residual(nls :: AbstractNLSModel, x :: AbstractVector)
-  Fx = zeros(nls_meta(nls).nequ)
+function residual(nls :: AbstractNLSModel, x :: AbstractVector{T}) where T <: Real
+  Fx = zeros(T, nls_meta(nls).nequ)
```
dpo: Is it necessary to parametrize? Couldn't we write `Fx = zeros(eltype(x), ...)`?

src/NLSModels.jl
```diff
@@ -268,11 +268,11 @@ end
 function obj(nls :: AbstractNLSModel, x :: AbstractVector)
   increment!(nls, :neval_obj)
   Fx = residual(nls, x)
-  return 0.5*dot(Fx, Fx)
+  return dot(Fx, Fx) / 2
```
dpo: Good catch!

src/NLSModels.jl
abelsiqueira: It works, but I actually don't know which is better.

src/NLSModels.jl
```diff
@@ -292,11 +292,11 @@ function objgrad!(nls :: AbstractNLSModel, x :: AbstractVector, g :: AbstractVec
...
   Fx = residual(nls, x)
   jtprod_residual!(nls, x, Fx, g)
-  return 0.5*dot(Fx, Fx), g
+  return dot(Fx, Fx) / 2, g
 end
 
-function hess(nls :: AbstractNLSModel, x :: AbstractVector; obj_weight ::
-              Float64 = 1.0, y :: AbstractVector = Float64[])
+function hess(nls :: AbstractNLSModel, x :: AbstractVector{T}; obj_weight ::
+              Real = one(T), y :: AbstractVector = T[]) where T<: Real
```
dpo: Here it may be necessary to parametrize.

src/NLSModels.jl
```diff
@@ -93,8 +93,8 @@ include("feasibility_residual.jl")
 
 Computes F(x), the residual at x.
 """
-function residual(nls :: AbstractNLSModel, x :: AbstractVector)
-  Fx = zeros(nls_meta(nls).nequ)
+function residual(nls :: AbstractNLSModel, x :: AbstractVector{T}) where T <: Real
+  Fx = zeros(T, nls_meta(nls).nequ)
   residual!(nls, x, Fx)
```
dpo: Here it seems to me that we make the assumption that problem functions (perhaps defined by the user) are able to work with various types. That's reasonable for Julia functions and ForwardDiff models (and hopefully JuMP soon). But what about others, such as CUTEstModels? I suppose it's ok to just let them raise an error?!

abelsiqueira: This heavily relies on `x`:
- If `residual!` works for `x` with `eltype=T`, then `residual` is well defined.
- If `residual!` doesn't for `x` with `eltype=T`, then `residual` doesn't work as well.

src/NLSModels.jl
abelsiqueira: I've made a new commit which uses only `eltype(x)`.
From what I tested:
- Using `x :: AbstractVector{T}` in some functions creates a cascade of using of `T` in other functions, like `SlackModel`, `Feasibility` stuff and `QNModels`.
- Using `T <: Real` also reduces the possibilities of what can be passed to function. We may want that, but it so then we would have to define it for every other function for consistency.

src/autodiff_nlsmodel.jl
```diff
@@ -165,11 +165,11 @@ function jtprod!(nls :: ADNLSModel, x :: AbstractVector, v :: AbstractVector, Jt
   return Jtv
 end
 
-function hess(nls :: ADNLSModel, x :: AbstractVector; obj_weight :: Real = 1.0, y :: AbstractVector = Float64[])
+function hess(nls :: ADNLSModel, x :: AbstractVector; obj_weight :: Real = one(eltype(x)), y :: AbstractVector = eltype(x)[]) where T <: Real
```
dpo: There is still a `where T` here.

test/multiple-precision.jl
```diff
@@ -0,0 +1,34 @@
+function multiple_precision()
+  @testset "Test multiple precision models" begin
+    for T = [Float64, Float32, Float16, BigFloat]
```
dpo: Maybe arrange them in order.

src/autodiff_model.jl
```diff
@@ -50,10 +50,10 @@ mutable struct ADNLPModel <: AbstractNLPModel
   c :: Function
 end
 
-function ADNLPModel(f::Function, x0::AbstractVector; y0::AbstractVector = Float64[],
-    lvar::AbstractVector = Float64[], uvar::AbstractVector = Float64[], lcon::AbstractVector = Float64[], ucon::AbstractVector = Float64[],
+function ADNLPModel(f::Function, x0::AbstractVector{T}; y0::AbstractVector = T[],
+    lvar::AbstractVector = T[], uvar::AbstractVector = T[], lcon::AbstractVector = T[], ucon::AbstractVector = T[],
     c::Function = (args...)->throw(NotImplementedError("cons")),
-    name::String = "Generic", lin::AbstractVector{Int}=Int[])
+    name::String = "Generic", lin::AbstractVector{Int}=Int[]) where T <: Real
```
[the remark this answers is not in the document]
abelsiqueira: I left in these "constructors" because they don't affect other functions (unlike the API), but I can change as well, what do you think?

src/autodiff_nlsmodel.jl
abelsiqueira: Thanks

src/autodiff_model.jl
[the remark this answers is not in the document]
dpo: Maybe it would be best to make everything uniform?!

src/autodiff_nlsmodel.jl
dpo: There were a few others.

abelsiqueira:
Thanks for the review. Updated.

dpo:
The rest seems good to me.

PR #165, opened 2019-02-07
Fix zenodo.json

abelsiqueira:
Adds contributors and license.


abelsiqueira:
This is how it looks like: https://zenodo.org/record/2559509

PR #167, opened 2019-02-14
Remove SimpleModel

abelsiqueira:
Just deletions


dpo:
Should this go to a separate repo SimpleNLPModels.jl?

abelsiqueira:
I think now for two reasons:
- I rather not maintain it;
- It's a very thin layer on top of just writing a specific model, such as `BROWNDEN <: AbstractNLPModel` and defining all api function for `nlp :: BROWNDEN`. The next PR shall introduce the problems in this format.

dpo:
I agree with you. Could the docs show a simple example? Perhaps they already do?!

abelsiqueira:
They news docs should, I'll add to my list, in case I forgot.

Issue #160, opened 2019-01-29
Coordinate Jacobian and Hessian improvements


abelsiqueira:
Considering the `foo_coord` functions which return `rows, cols, vals`.
We need
- A function call that simply gets the structure of the matrix - `rows, cols = foo_structure(nlp)`;
- Inplace functions to modify the vals - `foo_coord!(nlp, x, vals)`;

Consider using CUTEst as reference.

abelsiqueira:
@pjssilva

PR #172, opened 2019-02-20
[skip ci] Implement structure and inplace coordinate functions

abelsiqueira:
Rebasing into `coord-and-structure` to handle #160. After all PRs are merged into `coord-and-structure`, I'll open a PR which runs CI. **This doesn't run CI**.

Part 01 of 07.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -9,9 +9,10 @@ export AbstractNLPModelMeta, NLPModelMeta, AbstractNLPModel, Counters
 export reset!, sum_counters,
        obj, grad, grad!, objgrad, objgrad!, objcons, objcons!,
        cons, cons!, jth_con, jth_congrad, jth_congrad!, jth_sparse_congrad,
-       jac_coord, jac, jprod, jprod!, jtprod, jtprod!, jac_op, jac_op!,
+       jac_structure, jac_coord!, jac_coord,
+       jac, jprod, jprod!, jtprod, jtprod!, jac_op, jac_op!,
        jth_hprod, jth_hprod!, ghjvprod, ghjvprod!,
-       hess_coord, hess, hprod, hprod!, hess_op, hess_op!,
+       hess_structure, hess_coord!, hess_coord, hess, hprod, hprod!, hess_op, hess_op!,
        push!,
        varscale, lagscale, conscale,
        NotImplementedError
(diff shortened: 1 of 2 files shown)


src/NLPModels.jl
```diff
@@ -172,6 +173,19 @@ function objgrad!(nlp, x, g)
   return f, g
 end
 
+"""`(rows,cols) = jac_structure(nlp)`
```
dpo: I think the calling sequence should be on the second line, indented 4 spaces and with no backticks.

src/NLPModels.jl
```diff
@@ -172,6 +173,19 @@ function objgrad!(nlp, x, g)
...
+
+Returns the structure of the constraint's Jacobian in sparse coordinate format.
+"""
+jac_structure(:: AbstractNLPModel) = throw(NotImplementedError("jac_structure"))
+
+"""`(rows,cols,vals) = jac_coord!(nlp, x, rows, cols, vals)`
+
+Evaluate \$\\nabla c(x)\$, the constraint's Jacobian at `x` in sparse coordinate format,
```
dpo: How does the LaTeX show in the console? Is it readable?

src/NLSModels.jl
```diff
@@ -115,6 +117,25 @@ function jac_residual(nls :: AbstractNLSModel, x :: AbstractVector)
...
+  throw(NotImplementedError("jac_structure_residual"))
+end
+
+"""
+    (rows,cols,vals) = jac_coord_residual!(nls, x, rows, cols, vals)
+
+Computes the Jacobian of the residual at `x` in sparse coordinate format, rewriting
+`vals`. `rows` and `cols` are not rewritten.
+"""
+function jac_coord_residual!(nls :: AbstractNLSModel, x :: AbstractVector, rows :: AbstractVector{Int}, cols :: AbstractVector{Int}, vals :: AbstractVector)
```
dpo: Does this work? I thought `Vector{Int} <: AbstractVector{Int}` was `false`?!

src/NLPModels.jl
abelsiqueira: Here, but `\nabla c(x)` is colored differently, with a different color that `x`, `vals`, etc.
```
  (rows,cols,vals) = jac_coord!(nlp, x, rows, cols, vals)

  Evaluate \nabla c(x), the constraint's Jacobian at x in sparse coordinate format, rewriting vals.
  rows and cols are not rewritten.
```

src/NLSModels.jl
abelsiqueira: `Vector{Int} <: AbstractVector{Int}` is true, but `AbstractVector{Int32} <: AbstractVector{Int}` is false.

src/NLPModels.jl
abelsiqueira: Most docstrings in `NLPModels.jl` are this way. I'll open an issue to change them afterwards.

dpo: Ok

PR #173, opened 2019-02-20
[skip ci] Add tests

abelsiqueira:
Part 02 of 07 of #172

### DIFF
--- a/test/consistency.jl
+++ b/test/consistency.jl
@@ -84,6 +84,9 @@ function consistent_functions(nlps; rtol=1.0e-8, exclude=[])
     Hs = Vector{Any}(undef, N)
     for i = 1:N
       (I, J, V) = hess_coord(nlps[i], x)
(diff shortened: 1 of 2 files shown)


test/consistency.jl
```diff
@@ -84,6 +84,9 @@ function consistent_functions(nlps; rtol=1.0e-8, exclude=[])
     Hs = Vector{Any}(undef, N)
     for i = 1:N
       (I, J, V) = hess_coord(nlps[i], x)
+      IS, JS = hess_structure(nlps[i])
+      @test IS == I
+      @test JS == J
```
dpo: Should this be `.==` or `all(...)`?

abelsiqueira: `IS == I` is valid syntax.

dpo: You're right. It seems to contradict the dot notation in other places. I always do `all(IS .== I)`.

PR #176, opened 2019-02-20
[skip ci] Add coord-improvements to feasibility models, QN models and slack models

abelsiqueira:
Part 05 of 07 of #172


dpo:
Looks good!

PR #178, opened 2019-02-20
[skip ci] Simple memory tests for coord-improvements and add to docs

abelsiqueira:
Part 07 of 07 of #172


test/test_memory_of_coord.jl
```diff
@@ -0,0 +1,40 @@
+function test_memory_of_coord_of_nlp(nlp :: AbstractNLPModel)
+  n = nlp.meta.nvar
+  m = nlp.meta.nvar
+
+  x = 10 * [-(-1.0)^i for i = 1:n]
+  y = [-(-1.0)^i for i = 1:m]
+
+  # Hessian unconstrained test
+  rows, cols, vals = hess_coord(nlp, x)
+  al1 = @allocated rows, cols, vals = hess_coord(nlp, x)
```
dpo: Don't we know approximately how much it should be? `nnzh * (sizeof(Flotat64) + 2 * sizeof(Int)`?

abelsiqueira: I couldn't predict the size of the allocations. It allocates that amount, but also more. There are the headers, but also accessing vectors led to memory being used, and possibly something else I haven't figured out.
What I did find out is that `al2` is about `160` less than `al1`. I updated the testing checking for a decrease in memory used, and had to fix a few places that had wrong memory use. This has been rebased, also.

dpo: ok, thank you

PR #183, opened 2019-03-27
New methods to obtain Hessian structure in place

dpo:
When embedding an NLPModel into another (e.g., an augmented Lagrangian), it would be useful to be able to accumulate the structure into existing arrays.

### DIFF
--- a/src/NLSModels.jl
+++ b/src/NLSModels.jl
@@ -226,7 +226,7 @@ end
 """
     (rows,cols) = hess_structure_residual(nls)
 
-Returns the structure of the Lagrangian Hessian.
+Returns the structure of the residual Hessian.
 """
 function hess_structure_residual(nls :: AbstractNLSModel)
   throw(NotImplementedError("hess_structure_residual"))
--- a/src/NLSModels.jl
+++ b/src/NLSModels.jl
@@ -2,8 +2,8 @@ export AbstractNLSModel, nls_meta, NLSCounters, reset!,
        residual, residual!, jac_residual, jac_structure_residual, jac_coord_residual!,
        jac_coord_residual, jprod_residual,
        jprod_residual!, jtprod_residual, jtprod_residual!, jac_op_residual,
-       jac_op_residual!, hess_residual, hess_structure_residual, hess_coord_residual!,
-       hess_coord_residual, jth_hess_residual,
+       jac_op_residual!, hess_residual, hess_structure_residual, hess_structure_residual!,
+       hess_coord_residual!, hess_coord_residual, jth_hess_residual,
        hprod_residual, hprod_residual!, hess_op_residual, hess_op_residual!,
        NotImplementedError
 
@@ -229,7 +229,18 @@ end
 Returns the structure of the residual Hessian.
 """
 function hess_structure_residual(nls :: AbstractNLSModel)
-  throw(NotImplementedError("hess_structure_residual"))
+  rows = Vector{Int}(undef, nls.nls_meta.nnzh)
+  cols = Vector{Int}(undef, nls.nls_meta.nnzh)
+  hess_structure_residual!(nls, rows, cols)
+end
+
+"""
+    hess_structure_residual!(nls, rows, cols)
+
+Returns the structure of the residual Hessian in place.
+"""
+function hess_structure_residual!(nls :: AbstractNLSModel, rows :: AbstractVector{<: Integer}, cols :: AbstractVector{<: Integer})
+  throw(NotImplementedError("hess_structure_residual!"))
 end
 
 """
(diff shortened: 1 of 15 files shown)


src/NLSModels.jl
```diff
@@ -226,12 +226,21 @@ end
...
 function hess_structure_residual(nls :: AbstractNLSModel)
   throw(NotImplementedError("hess_structure_residual"))
 end
 
+"""
+    hess_structure_residual!(nls, rows, cols)
+
+Returns the structure of the residual Hessian in place.
+"""
+function hess_structure_residual(nls :: AbstractNLSModel, rows :: AbstractVector{<: Integer}, cols :: AbstractVector{<: Integer})
```
abelsiqueira: `!` missing

dpo: Fix coming up.

dpo:
> It would be nice if the default hess_structure calls hess_structure!, like grad! and other functions;

Done. I noticed other methods could do the same. That could be updated in a follow-up PR.

> I think the same idea can be applied to jac_structure and jac_structure_residual;

Yes, see above.

> The docs are failing because hess_structure! has to be added to docs/src/api.md;

Hm. It's in there though.

> The consistency tests have to add hess_structure! to the tests now;

The consistency tests call `hess_structure`, which calls `hess_structure!`, so I guess it's being tested. Is that sufficient?

> The existing methods need to implement this method.

Do you mean CUTEst, AMPL ,etc.? I opened a PR for CUTEst already.

> Maybe we should create a branch inplace-structure and PR to that until it becomes passing.

We can probably get it going one method at a time, don't you think?

dpo:
All green.

abelsiqueira:
Sorry for the delay.

> >The existing methods need to implement this method.
>
> Do you mean CUTEst, AMPL ,etc.? I opened a PR for CUTEst already.

Sorry, I meant "The existing models...", as in `ADNLPModel` and others inside `NLPModels.

You're right on the tests using `hess_structure!`, but it misses situations like this one, in which `ADNLPModel` implements `hess_structure` but not `hess_structure!`.

dpo:
I updated this but there remains some failures related to `feasibility_form_nls`. I'll submit a separate PR with other fixes.

src/feasibility_form_nls.jl
```diff
@@ -220,13 +220,15 @@ function jtprod!(nlp :: FeasibilityFormNLS, xr :: AbstractVector, v :: AbstractV
...
-  (Ic, Jc) = m > 0 ? hess_structure(nlp.internal) : (Int[], Int[])
-  I = [IF; Ic; (n+1:n+ne)]
-  J = [JF; Jc; (n+1:n+ne)]
-  return I, J
+  nnzhF = nlp.internal.nls_meta.nnzh
+  nnzhc = m > 0 ? nlp.internal.meta.nnzh : 0
+  I = 1:nnzhF
+  @views hess_structure_residual!(nlp.internal, rows[I], cols[I])
+  I = nnzhF+1:nnzhF+nnzhc
+  @views hess_structure!(nlp.internal, rows[I], cols[I])
```
abelsiqueira: Here we should check for `m > 0` and add the new residual Hessian.
```
if m > 0                                                                                  
  I = nnzhF+1:nnzhF+nnzhc                                                                 
  @views hess_structure!(nlp.internal, rows[I], cols[I])                                  
end                                                                                       
I = nnzhF+nnzhc+1:nnzhF+nnzhc+ne                                                          
rows[I] .= n+1:n+ne                                                                       
cols[I] .= n+1:n+ne 
```

src/feasibility_form_nls.jl
```diff
@@ -345,8 +347,9 @@ function hess_residual(nlp :: FeasibilityFormNLS, x :: AbstractVector, v :: Abst
   return spzeros(n, n)
 end
 
-function hess_structure_residual(nlp :: FeasibilityFormNLS)
-  return (Int[], Int[])
+function hess_structure_residual!(nlp :: FeasibilityFormNLS, rows :: AbstractVector{<: Integer}, cols :: AbstractVector{<: Integer})
+  increment!(nlp, :neval_hess_structure_residual)
```
abelsiqueira: There is no counter for `structure`, we can just remove this line.

abelsiqueira:
After these changes, it should pass, and you can merge then.

src/feasibility_form_nls.jl
dpo: I thought the check for `m > 0` was redundant. Wouldn't `nnzhc` be zero if `m` is zero?

abelsiqueira: Because of line
```
nnzhc = m > 0 ? nlp.internal.meta.nnzh : 0
```
we have `nnzhc = 0` if `m == 0`, but `nlp.internal.meta.nnzh` is not necessarily 0, because of the objective function of `nlp.internal`. This means that `hess_structure!` on `nlp.internal` will try to return `rows` and `cols` using `nlp.internal.meta.nnzh`, and in this case it will break.

abelsiqueira:
Thanks

dpo:
Thanks for the guidance on the NLS problems. This should be useful to your student who writes the augmented Lagrangian.

PR #184, opened 2019-03-30
Fix display math in docstrings.

xhu4:
Thank you for your great packages! I'm using NLPModels, CUTEst, NLPModelsJuMP and SolverTools for my research and they saved me a lot of time. I just realized some math display issues in your documentations (e.g. [here](http://juliasmoothoptimizers.github.io/NLPModels.jl/latest/api/#NLPModels.hess) and [here](http://juliasmoothoptimizers.github.io/NLPModels.jl/latest/models/#SlackModel-1)). So I tried to fix the problem.

Changes include:
- display math: fixed issues mentioned above,
- inline math: replaced `\$` by ` `` `,
- latex: replaced latex commands by unicode symbols for better docstring in REPL.

I'm pretty new to julia as well as collaborating on github. Let me know if I did anything stupid.
Thanks.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -3,6 +3,14 @@ __precompile__()
 
 module NLPModels
 
+# For documentation purpose
+const LAGRANGIAN_HESSION = raw"""
+```math
+∇²L(x,y) = σ ∇²f(x) + ∑_i yᵢ ∇²cᵢ(x),
+```
+with σ = obj_weight
+"""
+
 using LinearAlgebra, LinearOperators, Printf, SparseArrays, FastClosures
 
 export AbstractNLPModelMeta, NLPModelMeta, AbstractNLPModel, Counters
@@ -86,14 +94,14 @@ end
 # Methods to be overridden in other packages.
 """`f = obj(nlp, x)`
 
-Evaluate \$f(x)\$, the objective function of `nlp` at `x`.
+Evaluate ``f(x)``, the objective function of `nlp` at `x`.
 """
 obj(::AbstractNLPModel, ::AbstractVector) =
   throw(NotImplementedError("obj"))
 
 """`g = grad(nlp, x)`
 
-Evaluate \$\\nabla f(x)\$, the gradient of the objective function at `x`.
+Evaluate ``∇f(x)``, the gradient of the objective function at `x`.
 """
 function grad(nlp::AbstractNLPModel, x::AbstractVector)
   g = similar(x)
@@ -102,14 +110,14 @@ end
 
 """`g = grad!(nlp, x, g)`
 
-Evaluate \$\\nabla f(x)\$, the gradient of the objective function at `x` in place.
+Evaluate ``∇f(x)``, the gradient of the objective function at `x` in place.
 """
 grad!(::AbstractNLPModel, ::AbstractVector, ::AbstractVector) =
   throw(NotImplementedError("grad!"))
 
 """`c = cons(nlp, x)`
 
-Evaluate \$c(x)\$, the constraints at `x`.
+Evaluate ``c(x)``, the constraints at `x`.
 """
 function cons(nlp::AbstractNLPModel, x::AbstractVector)
   c = similar(x, nlp.meta.ncon)
@@ -118,7 +126,7 @@ end
 
 """`c = cons!(nlp, x, c)`
 
-Evaluate \$c(x)\$, the constraints at `x` in place.
+Evaluate ``c(x)``, the constraints at `x` in place.
 """
 cons!(::AbstractNLPModel, ::AbstractVector, ::AbstractVector) =
   throw(NotImplementedError("cons!"))
@@ -134,7 +142,7 @@ jth_sparse_congrad(::AbstractNLPModel, ::AbstractVector, ::Integer) =
 
 """`f, c = objcons(nlp, x)`
 
-Evaluate \$f(x)\$ and \$c(x)\$ at `x`.
+Evaluate ``f(x)`` and ``c(x)`` at `x`.
 """
 function objcons(nlp, x)
   f = obj(nlp, x)
@@ -144,7 +152,7 @@ end
 
 """`f = objcons!(nlp, x, c)`
 
-Evaluate \$f(x)\$ and \$c(x)\$ at `x`. `c` is overwritten with the value of \$c(x)\$.
+Evaluate ``f(x)`` and ``c(x)`` at `x`. `c` is overwritten with the value of ``c(x)``.
 """
 function objcons!(nlp, x, c)
   f = obj(nlp, x)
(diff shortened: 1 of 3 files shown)


src/NLPModels.jl
```diff
@@ -3,6 +3,14 @@ __precompile__()
 
 module NLPModels
 
+# For documentation purpose
+const LAGRANGIAN_HESSION = raw"""
+```math
+∇²L(x,y) = σ ∇²f(x) + ∑_i yᵢ ∇²cᵢ(x),
+```
+with σ = obj_weight
```
abelsiqueira: This needs quotes: `` `σ = obj_weight` ``

src/slack_model.jl
```diff
@@ -5,32 +5,47 @@ export SlackModel, SlackNLSModel,
...
+                  c_L &≤& c(x) &≤& c_U,\\
+                   l  &≤&   x  &≤&  u,
+                \end{array}
+\end{align}
+```
 
 the new model appears to the user as
 
-\\\\[ \\min f(X)  \\mbox{ s. t. }  g(X) = 0 \\mbox{ and } L \\leq X \\leq U. \\\\]
+```math
```
abelsiqueira: I believe `\begin{align}` is missing here, and possibly the `array`, so that it's similar to the expression above.

src/slack_model.jl
```diff
@@ -5,32 +5,47 @@ export SlackModel, SlackNLSModel,
...
 
-The unknowns \$X = (x, s)\$ contain the original variables and slack variables
-\$s\$. The latter are such that the new model has the general form
+The unknowns ``X = (x, s)`` contain the original variables and slack variables
+``s``. The latter are such that the new model has the general form
 
-\\\\[ \\min f(x)  \\mbox{ s. t. }  c(x) - s = 0, c_L \\leq s \\leq c_U \\mbox{ and }
-\\ell \\leq x \\leq u, \\\\]
+```math
+&\min_Xx     & &f(x)\\
```
abelsiqueira: Same here

src/slack_model.jl
```diff
@@ -5,32 +5,47 @@ export SlackModel, SlackNLSModel,
...
+&\text{s.t.} & &c(x) - s = 0,\\
+&            & &c_L ≤ x ≤ c_U,\\
+&            & & l  ≤ x ≤ u.
+```
 
 although no slack variables are introduced for equality constraints.
 
 The slack variables are implicitly ordered as [s(low), s(upp), s(rng)], where
```
abelsiqueira: Add quotes between `[s(low), s(upp), s(rng)]`

src/NLPModels.jl
```diff
@@ -3,6 +3,14 @@ __precompile__()
 
 module NLPModels
 
+# For documentation purpose
+const LAGRANGIAN_HESSION = raw"""
```
abelsiqueira: Should be `HESSIAN`.

src/slack_model.jl
xhu4: I can't see how to align the equation `g(x) = 0` with chained inequality `L ≤ X ≤ U` in `array`. But with/without `array` does create some discrepancy. The `array` environment is creating extra spaces between columns:
array
 and I cannot find a way to remove that yet. (I tried `\setlength\arraycolsep{1pt}` and `\begin{array}[t]{c@{\hspace{1pt}}c...}`, both work with Latex but not with Documenter.jl). 

A workaround is to use `alignedat`, which results in something like this:
alignedat
I guess this is the best we can have.

src/NLPModels.jl
xhu4: :joy:Nice catch

src/slack_model.jl
```diff
@@ -5,32 +5,47 @@ export SlackModel, SlackNLSModel,
...
+&            & &L ≤ X ≤ U.
+```
 
-The unknowns \$X = (x, s)\$ contain the original variables and slack variables
-\$s\$. The latter are such that the new model has the general form
+The unknowns ``X = (x, s)`` contain the original variables and slack variables
+``s``. The latter are such that the new model has the general form
 
-\\\\[ \\min f(x)  \\mbox{ s. t. }  c(x) - s = 0, c_L \\leq s \\leq c_U \\mbox{ and }
-\\ell \\leq x \\leq u, \\\\]
```
xhu4: I changed all `\ell` to plain `l`. Is there any reason to prefer `\ell` over `l`? I just saw some other `\ell`s in some markdown files. Maybe we should keep accordance?

abelsiqueira: Lowercase L looks like 1, and though it's usually clear from context, we prefer `\ell`.

src/slack_model.jl
abelsiqueira: I think we can use the same format as in the docs (https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/docs/src/index.md), i.e., simple `align`.

src/slack_model.jl
xhu4: I see. I changed back all the `\ell`s in the new commit.

src/slack_model.jl
xhu4: I'll leave it with `alignedat` for now. I think it's slightly better than just `align`. Let me know if you prefer `align`.

abelsiqueira: Sorry for the delay. What I meant was changing to something simpler overall, like this:
```
\begin{align*}                                                                              
       \min_x \quad & f(x) \\                                                               
\mathrm{s.t.} \quad & c_L ≤ c(x) ≤ c_U,\\                                                   
                    &   ℓ ≤   x  ≤ u,                                                      
\end{align*}                                                                                
```
which uses only a simple align, and is also easy to read on the help.

xhu4: Less is more. Got it. Do I need to care about the display maths in markdown files?

abelsiqueira: Sorry, what do you mean?

xhu4: Forget it. I was thinking about updating the maths in docs/*.md like realigning and using Unicode. But since markdown files will not appear in REPL as source code, I guess we can just leave them alone.

src/slack_model.jl
```diff
@@ -5,32 +5,49 @@ export SlackModel, SlackNLSModel,
...
+```
+
+The unknowns ``X = (x, s)`` contain the original variables and slack variables
+``s``. The latter are such that the new model has the general form
+
+```math
+\begin{align*}
+       \min_x \quad & f(x)\\
+\mathrm{s.t.} \quad & c(x) - s = 0,\\
+                    & c_L ≤ x ≤ c_U,\\
```
abelsiqueira: This should be `s` instead of `x`: `c_L ≤ s ≤ c_U`

abelsiqueira:
Thanks, @xhu4.

xhu4:
My pleasure.

PR #185, opened 2019-04-02
fixes to feasibility_form_nls

dpo:
Its seems to me that the Hessian in feasibility_form_nls didn't take the objective ¹/₂‖r‖² into account.

### DIFF
--- a/src/feasibility_form_nls.jl
+++ b/src/feasibility_form_nls.jl
@@ -239,10 +239,10 @@ function hess_coord!(nlp :: FeasibilityFormNLS, xr :: AbstractVector, rows :: Ab
   y2 = @view y[ne+1:ne+m]
   I = 1:nnzhF
   @views hess_coord_residual!(nlp.internal, x, y1, rows[I], cols[I], vals[I])
-  if m > 0
-    I = nnzhF+1:nnzhF+nnzhc
-    @views hess_coord!(nlp.internal, x, rows[I], cols[I], vals[I], obj_weight=0.0, y=y2)
-  end
+  I = nnzhF+1:nnzhF+nnzhc
+  @views hess_coord!(nlp.internal, x, rows[I], cols[I], vals[I], obj_weight=0.0, y=y2)
+  @views rows[nnzhF + nnzhc + 1 : nnzhF + nnzhc + ne] = n - ne + 1 : n  # Hessian of the residual ¹/₂‖r‖²
+  @views cols[nnzhF + nnzhc + 1 : nnzhF + nnzhc + ne] = n - ne + 1 : n
   vals[nnzhF+nnzhc+1:nnzhF+nnzhc+ne] .= obj_weight
   return rows, cols, vals
 end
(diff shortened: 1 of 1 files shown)


src/feasibility_form_nls.jl
```diff
@@ -239,10 +239,10 @@ function hess_coord!(nlp :: FeasibilityFormNLS, xr :: AbstractVector, rows :: Ab
...
   I = 1:nnzhF
   @views hess_coord_residual!(nlp.internal, x, y1, rows[I], cols[I], vals[I])
-  if m > 0
-    I = nnzhF+1:nnzhF+nnzhc
-    @views hess_coord!(nlp.internal, x, rows[I], cols[I], vals[I], obj_weight=0.0, y=y2)
-  end
+  I = nnzhF+1:nnzhF+nnzhc
+  @views hess_coord!(nlp.internal, x, rows[I], cols[I], vals[I], obj_weight=0.0, y=y2)
+  @views rows[nnzhF + nnzhc + 1 : nnzhF + nnzhc + ne] = n - ne + 1 : n  # Hessian of the residual ¹/₂‖r‖²
+  @views cols[nnzhF + nnzhc + 1 : nnzhF + nnzhc + ne] = n - ne + 1 : n
```
abelsiqueira: We don't change `rows` and `cols`.

Below, line 246, should be the Hessian of `¹/₂‖r‖²` times `obj_weight`.

src/feasibility_form_nls.jl
```diff
@@ -339,10 +339,15 @@ function jtprod_residual!(nlp :: FeasibilityFormNLS, x :: AbstractVector, v :: A
...
+# the Hessian of the residual is
+# [ 0  0 ]  } n
+# [ 0  I ]  } m
 function hess_residual(nlp :: FeasibilityFormNLS, x :: AbstractVector, v :: AbstractVector)
   increment!(nlp, :neval_hess_residual)
   n = nlp.meta.nvar
-  return spzeros(n, n)
+  m = nlp.internal.nls_meta.nequ
+  rows = collect((n-m+1):n)
+  return sparse(rows, rows, ones(eltype(x), m), n, n)
```
abelsiqueira: The residual function, in this case, is `F(x,r) = r`, so the Hessian should really be 0.

abelsiqueira:
The residual functions use `F` from `0.5|F(x)|^2`, so `hess_residual(nlp, x, v)` is the sum of Hessians of each residual `F_i` times `v_i`. `FeasibilityFormNLS` takes

`F(x)` from residual and `c(x)` from constraints and defines `F2(x,r) = r` and `c2(x,r) = [F(x) - r; c(x)]`. So the new residual is linear, and the residual-Hessian related things are 0. `hess` and `hprod` should still have contributions from `F2`.

src/feasibility_form_nls.jl
abelsiqueira: A problem that may happen here is that for `NLS` problems some `hess` functions may be undefined due to the `J'*J` term. So calling `hess_coord!` here may lead to a `NotImplementedError`.

src/feasibility_form_nls.jl
dpo: Maybe I'm confused as to what `hess_residual` is supposed to return. Is it the sum of the Hessians of the individual residuals?

dpo:
Ok, I guess this is pointless then. Sorry for the confusion.

src/feasibility_form_nls.jl
abelsiqueira: Yes.

abelsiqueira:
No problem. Residual functions are really confusing right now, hopefully it'll be better with NLPModels 0.10.0 (TBA).

PR #188, opened 2019-07-05
Bang methods

dpo:
Better definition of default methods, where non-`!`methods call the corresponding `!` methods.


dpo:
The build failure is related to the docs, but I don't understand why it fails.

abelsiqueira:
Thanks a lot!

PR #189, opened 2019-07-05
ADNLPModel should return a dense Hessian

dpo:


abelsiqueira:
I saw you added the WIP tag, do you plan to make more modifications?

dpo:
Yes, there are a couple of related changes I'd like to push here.

dpo:
Should be good now.

PR #190, opened 2019-07-06
Simplify methods

dpo:
#188 will allow to remove quite a bit of code here, in CUTEst and in AmplNLReader.


abelsiqueira:
Very nice, thanks.
I've created release 0.8.1 (https://github.com/JuliaRegistries/General/pull/1840) contained #188, should be good to go soon.

dpo:
Thanks! We should try and do the same with NLS models. As soon as 0.8.1 ships, I'll rerun similar PRs for CUTEst and AmplNLReader.

PR #205, opened 2019-08-23
Change coverage in travis to run only for one build

abelsiqueira:
Might be interesting to have `isunix` instead of `islinux` for some packages, but NLPModels doesn't have any specific OS code. 
On Windows the coverage submission sends wrong filenames, which can't be explored (`src\NLPModels.jl` instead of `src/NLPModels.jl`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,8 @@ jobs:
       after_success: skip
 
 after_success:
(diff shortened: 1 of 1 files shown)


dpo:
Good idea, thanks!

PR #207, opened 2019-08-25
Test packages depending on NLPModels with a submitted pull request

abelsiqueira:
If this works, after the usual tests pass, a new stage testing NLPModels-dependent packages will start and a comment will be created in the repo showing what passed.

### DIFF
--- a/test/test-breakage.jl
+++ b/test/test-breakage.jl
@@ -1,0 +1,103 @@
+using Git, GitHub, Pkg
+
+"""
+    test_breakage()
+
+Downloads each package and check if tests pass using the master of NLPModels.jl.
+Packages are checked in two situations:
+
+- Tagged version
+- master branch
+
+If any tagged version breaks then the next version should be a major update or
+the broken package has bugs.
+"""
+function test_breakage()
+  packages = ["AmplNLReader", "CaNNOLeS", "CUTEst", "QuadraticModels", "NLPModelsIpopt",
+              "NLPModelsJuMP", "NLPModelsKnitro", "SolverTools"]
+
+  msg(s) = printstyled(s, color=:red, bold=true)
+  msgn(s) = printstyled(s * "\n", color=:red, bold=true)
+
+  passing = fill(false, length(packages))
+  passing_master = fill(false, length(packages))
+  tagged = fill("", length(packages))
+
+  master_fail = "![](https://img.shields.io/badge/master-Fail-red)"
+  master_pass = "![](https://img.shields.io/badge/master-Pass-green)"
+  version_fail(n) = "![](https://img.shields.io/badge/$n-Fail-red)"
+  version_pass(n) = "![](https://img.shields.io/badge/$n-Pass-green)"
+
+  thispath = pwd()
+  mkpath("test-breakage")
+  try
+    for (i,package) in enumerate(packages)
+      msgn("#"^100)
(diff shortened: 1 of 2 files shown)


dpo:
I just looked at the log of the AmplNLReader install and saw
```
Installed NLPModels ─ v0.8.3
```
Don't we want to use the version of NLPModels in the current pull request instead? So somehow, we should go into `TRAVIS_BUILD_DIR` and do a `pkg"dev" .`

Also, this scheme would break if a pull requested updated the NLPModels version number and that version number were out of range for certain dependent packages, right?!

abelsiqueira:
I do `pkg"dev ../.."`. The result is [here](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/576476792#L351)

You're right the tests would fail in that situation. We'll have to limit the script being run to PRs so it doesn't break during master and tag builds, but otherwise, I don't think there is much to do.

My main concern with the script is to find out whether to update a minor or major version, so it will usually break, but if I did things right, it won't be marked as a failed test by travis.

abelsiqueira:
Just found out that "Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code." therefore, my idea does not work.

abelsiqueira:
Actually, just the fun part doesn't work, but still.

dpo:
What encrypted environment variables are we talking about?

abelsiqueira:
`GITHUB_AUTH`. I want to comment on the pull request with the summary, like this:
https://github.com/abelsiqueira/NLPModels.jl/pull/2#issuecomment-524676377

abelsiqueira:
I think that we should:

- Run the script locally, which should upload the results using the local user's account;
- Set the script to run on `master` instead of PRs, so we still know what will break automatically, but only after the fact;
- In the future, invest more time (and possibly money) on the subject, looking into GitHub Apps.

abelsiqueira:
:robot: Testing breakage of this commit
- AmplNLReader  
- CaNNOLeS  
- CUTEst  
- QuadraticModels  
- NLPModelsIpopt  
- NLPModelsJuMP  
- NLPModelsKnitro  
- SolverTools

dpo:
It's got to be possible since services such a the registration bot do it. Perhaps by creating a new user and storing that user's GITHUB_AUTH in the Travis config?!

test/test-breakage.jl
```diff
@@ -0,0 +1,103 @@
...
+      if p.merge_commit_sha == ENV["TRAVIS_COMMIT"]
+        pr = p
+      end
+    end
+
+    output = ":robot: Testing breakage of this commit\n"
+    for (i,package) in enumerate(packages)
+      output *= "- $package "
+      output *= (passing_master[i] ? master_pass : master_fail) * " "
+      output *= (passing[i] ? version_pass(tagged[i]) : version_fail(tagged[i])) * "\n"
```
dpo: Would it look ok as a Markdown table?

dpo:
Let's make this work. It's absolutely awesome!

abelsiqueira:
I believe the registration bot is a GitHub App, and it should be possible to do what we want, it's just a matter of time.
Creating a new user won't work, because Travis blocks `GITHUB_AUTH` from being used. I tried with my GITHUB_AUTH, that's how it works on my repository.

PR #210, opened 2019-08-27
Remove unnecessary exports and non-inplace functions

abelsiqueira:


abelsiqueira:
:robot: Testing breakage of this commit

| Package Name | master | Tagged |
|--|--|--|
| AmplNLReader |  |  |
| CaNNOLeS |  |  |
| CUTEst |  |  |
| QuadraticModels |  |  |
| NLPModelsIpopt |  |  |
| NLPModelsJuMP |  |  |
| SolverTools |  |  |

src/NLSModels.jl
```diff
@@ -340,60 +336,16 @@ function obj(nls :: AbstractNLSModel, x :: AbstractVector)
...
-end
-
 function objgrad!(nls :: AbstractNLSModel, x :: AbstractVector, g :: AbstractVector)
   increment!(nls, :neval_obj)
   increment!(nls, :neval_grad)
   Fx = residual(nls, x)
   jtprod_residual!(nls, x, Fx, g)
   return dot(Fx, Fx) / 2, g
 end
-
```
abelsiqueira: Hessian functions for a general NLS with constraints cannot be defined, so I removed these and passed the responsibility to specific NLS models.

test/consistency.jl
```diff
@@ -16,6 +16,9 @@ function consistent_counters(nlps)
   V = zeros(Int, N)
   for field in fieldnames(Counters)
     V = [eval(field)(nlp) for nlp in nlps]
+    if !all(V .== V[1])
```
abelsiqueira: Added these because these failures were not informative

abelsiqueira:
:robot: Testing breakage of this commit

| Package Name | master | Tagged |
|--|--|--|
| AmplNLReader |  |  |
| CaNNOLeS |  |  |
| CUTEst |  |  |
| QuadraticModels |  |  |
| NLPModelsIpopt |  |  |
| NLPModelsJuMP |  |  |
| SolverTools |  |  |

src/NLSModels.jl
dpo: Wouldn't it be sufficient to also define `hess` based on `hess_coord` and ask the user to implement `hess_coord!`?

src/autodiff_model.jl
```diff
@@ -100,11 +99,6 @@ function cons!(nlp :: ADNLPModel, x :: AbstractVector, c :: AbstractVector)
   return c
 end
 
-function jac(nlp :: ADNLPModel, x :: AbstractVector)
-  increment!(nlp, :neval_jac)
-  return ForwardDiff.jacobian(nlp.c, x)
-end
-
```
dpo: Do we want the Jacobian to be sparse or dense here?

src/NLSModels.jl
abelsiqueira: We'd need `nnzj` defined to populate `rows` and `cols` and that is model dependent also.

src/autodiff_model.jl
abelsiqueira: We don't have sparse matrices with ForwardDiff, unfortunately.

dpo: What I mean is that if you delete `jac()` here, then, the default `jac()` will be used and it creates a sparse matrix.

src/NLSModels.jl
dpo: We can figure out the sparsity pattern of J'J, and therefore that of H+J'J.

src/autodiff_model.jl
abelsiqueira: Ah sorry, I see what you mean. I think we should return this then, so it's exactly what ForwardDiff gives.

src/NLSModels.jl
abelsiqueira: Do you mean we assume that `rows` and `cols` will be well-defined? For instance, in `LLSModel` I assumed that computing the coordinate Hessian would be too expensive and used `nnzh = 0`. Instead, we should compute `nnzh` in the constructor of `LLSModel` using the same computations made here?

Going back to my original statement, what I meant was that if `y` is given, we don't know what to do, because we don't have a way to compute the Hessian of the constraints.

src/autodiff_model.jl
abelsiqueira: Fixed.

abelsiqueira:
:robot: Testing breakage of this commit

| Package Name | master | Tagged |
|--|--|--|
| AmplNLReader |  |  |
| CaNNOLeS |  |  |
| CUTEst |  |  |
| QuadraticModels |  |  |
| NLPModelsIpopt |  |  |
| NLPModelsJuMP |  |  |
| SolverTools |  |  |

abelsiqueira:
:robot: Testing breakage of this commit

| Package Name | master | Tagged |
|--|--|--|
| AmplNLReader |  |  |
| CaNNOLeS |  |  |
| CUTEst |  |  |
| QuadraticModels |  |  |
| NLPModelsIpopt |  |  |
| NLPModelsJuMP |  |  |
| SolverTools |  |  |

abelsiqueira:
@dpo, I fixed the breakage. This should be ready again.

abelsiqueira:
:robot: Testing breakage of this commit

| Package Name | master | Tagged |
|--|--|--|
| AmplNLReader |  |  |
| CaNNOLeS |  |  |
| CUTEst |  |  |
| QuadraticModels |  |  |
| NLPModelsIpopt |  |  |
| NLPModelsJuMP |  |  |
| SolverTools |  |  |

dpo:
Thank you.

PR #218, opened 2019-09-26
Test jac_structure!

abelsiqueira:


src/qn_model.jl
```diff
@@ -57,6 +57,7 @@ for meth in (:jprod!, :jtprod!)
   @eval $meth(nlp :: QuasiNewtonModel, x :: AbstractVector, y :: AbstractVector, z :: AbstractVector) = $meth(nlp.model, x, y, z)
 end
 jac_structure(nlp :: QuasiNewtonModel) = jac_structure(nlp.model)
```
dpo: Maybe `jac_structure()` is no longer necessary?!

dpo:
The breakage job fails with `ERROR: LoadError: KeyError: key "GITHUB_AUTH" not found`.

abelsiqueira:
Rebased and working now. This PR is from outside JSO, so there is no breakage test. But everything is already broken from the last PR, so it shouldn't make a difference.

dpo:
Obrigado.

PR #221, opened 2019-11-13
Version 0.10.0

abelsiqueira:
New release?


dpo:
Do we first need to limit the other packages?

PR #239, opened 2020-02-13
Update docs/make.jl for Documenter 0.24

amontoison:


dpo:
What is this for? There is no explanation.

amontoison:
I'm working on an MathOptNLPModel that create an NLPModel from a JuMP model.
I interfaced MOI functions available for an MOI.AbstractNLPEvaluator with generic NLPModel functions.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/nlp.jl

But `objective_expr` and `constraint_expr` are the only functions that are not available in NLPModel.
I think that it can be useful to get expression tree of the objective function / constraints in some cases.

dpo:
It will be specific to MathOptNLPModels though, right? No other model can return an expression?!

amontoison:
Yes, to my knowledge it's the only one that has this feature.

dpo:
Is it useful to define those methods here then? Shouldn't they just be in NLPModelsJuMP?

amontoison:
Yes, you're right.
I will just update Documenter with this pull request.

docs/make.jl
```diff
@@ -19,8 +16,4 @@ makedocs(
           ]
 )
 
-deploydocs(deps = nothing, make = nothing,
```
amontoison: Other keywords used they value by default, so we can remove them.
I'm still trying to find why, after deployment, links to source don't work...

dpo:
Thank you.

PR #245, opened 2020-02-22
Create prod functions that accept structure

abelsiqueira:
cf @dpo, @amontoison 

Create/modify:
1. `jprod!(nlp, rows, cols, vals, v, Jv)` and `jtprod!` - well defined
2. `jprod!(nlp, x, rows, cols, v, Jv)` and `jtprod!` - defaults to `jprod!(nlp, x, v, Jv)`
3. `hprod!(nlp, rows, cols, vals, v, Hv)` (this works for both unc and con) - well defined
4. `hprod!(nlp, x, rows, cols, v, Hv)` - defaults to `hprod!(nlp, x, v, Hv)`
5. `hprod!(nlp, x, y, rows, cols, v, Hv)` - defaults to `hprod!(nlp, x, y, v, Hv)`
6. `jac_op!(nlp, rows, cols, vals, Jv, Jtv)` - uses 1
7. `jac_op!(nlp, x, rows, cols, Jv, Jtv)` - call `vals = jac_coord` and uses 6
8. `jac_op!(nlp, x, Jv, Jtv)` - modified to call `rows, cols = jac_structure` and `vals = jac_coord` and uses 6
9. `hess_op!` has same cases as 6, 7, 8
10. `decrement!` is implemented for use in 7.

PR #247, opened 2020-02-24
Create guidelines in docs

abelsiqueira:
Non-exhaustive, and can be updated after #245.

### DIFF
--- a/docs/src/guidelines.md
+++ b/docs/src/guidelines.md
@@ -4,9 +4,9 @@ These are guidelines for the creation of models using NLPModels to help keeping 
 
 Table of contents:
 - [Bare minimum](@ref bare-minimum)
-- [Testing your model](@ref testing-your-model)
 - [Expected behaviour](@ref expected-behaviour)
 - [Advanced counters](@ref advanced-counters)
+- [Advanced tests](@ref advanced-tests)
 
 ## [Bare minimum](@id bare-minimum)
 
@@ -72,7 +72,47 @@ The following functions should be defined:
   - `hess_coord!(nlp, x, y, hvals; obj_weight=1)`
   - `hprod!(nlp, x, y, v, Hv; obj_weight=1)`
 
-## [Testing your model](@id testing-your-model)
+## [Expected behaviour](@id expected-behaviour)
+
+The following is a non-exhaustive list of expected behaviour for methods.
+
+- All in place methods should also return the modified vectors.
+- Vector inputs should have the correct size. If necessary, the user should pass them using views or slices.
+- The triplet format does not assume order nor uniqueness.
+
+## [Advanced counters](@id advanced-counters)
+
+If a model does not implement `counters`, then it needs to define
+- `neval_xxx(nlp)` - get field `xxx` of `Counters`
+- `reset!(nlp)` - resetting all counters
+- `increment!(nlp, s)` - increment counter `s`
+For instance
+```julia
+for counter in fieldnames(Counters)
+  @eval begin
+    $counter(nlp :: MyModel) = SOMETHING
+  end
+end
+function reset!(nlp :: MyModel)
+  RESET COUNTERS
+end
+function increment!(nlp :: MyModel, s :: Symbol)
+  INCREMENT COUNTER s
+end
+```
+One example of such model is the `SlackModel`, which stores an internal `model :: AbstractNLPModel`, thus defining
+```julia
+$counter(nlp :: SlackModel) = $counter(nlp.model)
+reset!(nlp :: SlackModel) = reset!(nlp.model)
+increment!(nlp :: SlackModel, s :: Symbol) = increment!(nlp.model, s)
+```
+This construction can be replicated calling the macro `@default_counters Model inner`.
+In the case of SlackModel, the equivalent call is
+```julia
+@default_counters SlackModel model
+```
+
+## [Advanced tests](@id advanced-tests)
 
 To test your model, in addition to writing specific test functions, it is also advised to write consistency checks.
 If your model can implement general problems, you can use the 6 problems in our `test/problems` folder implemented both as `ADNLPModel` and by explicitly defining these problem as models.
@@ -114,42 +154,3 @@ consistent_nlps([rnlp, manual])
 ```
 The complete example is available in the repository [RegularizationModel.jl](https://github.com/JuliaSmoothOptimizers/RegularizationModel.jl).
 
-## [Expected behaviour](@id expected-behaviour)
-
-The following is a non-exhaustive list of expected behaviour for methods.
-
-- All in place methods should also return the modified vectors.
-- Vector inputs should have the correct size. If necessary, the user should pass them using views or slices.
-- The triplet format does not assume order nor uniqueness.
-
-## [Advanced counters](@id advanced-counters)
-
-If a model does not implement `counters`, then it needs to define
-- `neval_xxx(nlp)` - get field `xxx` of `Counters`
-- `reset!(nlp)` - resetting all counters
-- `increment!(nlp, s)` - increment counter `s`
-For instance
-```julia
-for counter in fieldnames(Counters)
-  @eval begin
-    $counter(nlp :: MyModel) = SOMETHING
-  end
-end
-function reset!(nlp :: MyModel)
-  RESET COUNTERS
-end
-function increment!(nlp :: MyModel, s :: Symbol)
-  INCREMENT COUNTER s
-end
-```
-One example of such model is the `SlackModel`, which stores an internal `model :: AbstractNLPModel`, thus defining
-```julia
-$counter(nlp :: SlackModel) = $counter(nlp.model)
-reset!(nlp :: SlackModel) = reset!(nlp.model)
-increment!(nlp :: SlackModel, s :: Symbol) = increment!(nlp.model, s)
-```
-This construction can be replicated calling the macro `@default_counters Model inner`.
-In the case of SlackModel, the equivalent call is
-```julia
-@default_counters SlackModel model
-```
--- a/docs/src/guidelines.md
+++ b/docs/src/guidelines.md
@@ -2,21 +2,17 @@
 
 These are guidelines for the creation of models using NLPModels to help keeping the models uniform, and for future reference in the creation of solvers.
 
-## Bare minimum
+Table of contents:
+- [Bare minimum](@ref bare-minimum)
+- [Testing your model](@ref testing-your-model)
+- [Expected behaviour](@ref expected-behaviour)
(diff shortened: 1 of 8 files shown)


docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
+# [Guidelines for creating models](@id Guidelines)
+
+These are guidelines for the creation of models using NLPModels to help keeping the models uniform, and for future reference in the creation of solvers.
+
+## Bare minimum
+
+Your model should derive from `AbstractNLPModel` or some other abstract class derived from it.
+It is mandatory that it has a `meta :: NLPModelMeta` field, storing all the relevant problem information.
```
dpo: has -> have (this is the subjunctive voice).

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+The minimum information that should be set is
+
+- `nvar`: Number of variables
```
dpo: Isn't `nvar the minimum information? The rest has default values.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `x0`: Starting point (defaults to `zeros`)
+- `lvar`, `uvar`: Bounds on the variables (default to `(-∞,∞)`)
+- `ncon`: Number of constraints (defaults to `0`)
+- `lcon`, `ucon`: Bounds on the constraints (default to `(-∞,∞)`)
+- `nnzh`: The length of the vectors used to store a triangle of the Hessian matrix in triplet format (defaults to `nvar * (nvar + 1) / 2`
```
dpo: No need to say "matrix" here.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `nnzj`: The length of the vectors used to store the Jacobian matrix in triplet format (default to `nvar * ncon`)
```
dpo: Or here.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+
+The model also need to provide `Counters` information, the easiest way being defining `counters :: Counters`.
```
dpo: needs ... (full stop). The easiest way is to define

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+For instance:
+```julia
+mutable struct MyModel <: AbstractNLPModel
+  meta :: NLPModelMeta
+  counters :: Counters
+end
```
dpo: the part about counters and this example should be moved above what needs to go into meta.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+If a model does not implement `counters`, then it needs to define `neval_xxx` functions, such as in the following macro:
+```julia
+for counter in fieldnames(Counters)
+  @eval begin
+    $counter(nlp :: MyModel) = SOMETHING
+  end
+end
```
dpo: Just a thought here. We could easily define a macro `@default_counters` (or something similar) so that the line
```julia
@default_counters MyModel inner_model
```
expands to
```julia
for counter in fieldnames(Counters)
   @eval begin
     $counter(nlp :: MyModel) = $counter(nlp.inner_model)
   end
 end
```

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+```
+One example of such model is the `SlackModel`, which stores an internal `model :: AbstractNLPModel`, thus defining
+```julia
+$counter(nlp :: SlackModel) = $counter(nlp.model)
+```
+
+There are about 30 functions in the NLPModels API, and a few with more than one signature, so it can be scary to implement them all.
```
dpo: I'm not sure the last part is helpful here.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+Luckily, there are some functions with default definition to use other functions (e.g., `grad` is defined through `grad!` by default).
```
dpo: Luckily, many have a default implementation.

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+We compile here the list of functions that should be implemented for a complete API.
+
+Here, the following notation apply:
+- `nlp` is your model `MyModel <: AbstractNLPModel`
```
dpo: model -> instance of

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+end
+```
+One example of such model is the `SlackModel`, which stores an internal `model :: AbstractNLPModel`, thus defining
+```julia
+$counter(nlp :: SlackModel) = $counter(nlp.model)
+```
+
+There are about 30 functions in the NLPModels API, and a few with more than one signature, so it can be scary to implement them all.
+Luckily, there are some functions with default definition to use other functions (e.g., `grad` is defined through `grad!` by default).
+We compile here the list of functions that should be implemented for a complete API.
```
dpo: compile -> collect

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+
+Here, the following notation apply:
+- `nlp` is your model `MyModel <: AbstractNLPModel`
+- `x` is the point where the function is evaluated
+- `y` is the vector of Lagrange multipliers
```
dpo: (for constrained problems only)

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `g` is the gradient vector
+- `H` is the Hessian matrix of the objective or Lagrangian
```
dpo: remove "matrix"

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `hrows`, `hcols`, and `hvals` are vectors storing the triplet form of the Hessian
+- `c` is the constraints vector
```
dpo: vector of constraints

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `J` is the Jacobian matrix of the constraints
```
dpo: no "matrix"

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+- `jrows`, `jcols`, and `jvals` are vectors storing the triplet form of the Jacobian
+- `v` is a vector of appropriate dimensions, generally used for matrix-vector products
+- `Jv`, `Jtv`, `Hv` are vectors of appropriate dimensions, storing the result of matrix-vector products
```
dpo: I think we should say "operator-vector products".

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+  - `jac_structure!(nlp, jrows, jcols)`
+  - `jac_coord!(nlp, x, jvals)`
+  - `jprod!(nlp, x, v, Jv)`
+  - `jtprod!(nlp, x, v, Jtv)`
+  - `hess_coord!(nlp, x, y, hvals; obj_weight=1)`
+  - `hprod!(nlp, x, y, v, Hv; obj_weight=1)`
+
+## Testing your model
+
+To test your model, in addition to writing specific testing functions, it is also advised to write consistency checks.
```
dpo: test functions

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+The simplest way to use these would be something like
+```julia
+for problem in ["BROWNDEN", "HS5", "HS6", "HS10", "HS11", "HS14"]
+  @printf("Checking problem %-20s", problem)
+  nlp_ad = eval(Meta.parse(lowercase(problem) * "_autodiff"))() # e.g. hs5_autodiff()
+  nlp_man = eval(Meta.parse(problem))() # e.g. HS5()
+  nlp_your = ...
+  nlps = [nlp_ad, nlp_man, nlp_your]
+  consistent_nlps(nlps)
+end
```
dpo: This only applies to "modeling languages", right? If I write a subtype of AbstractNLPModels to represent an augmented-Lagrangian subproblem, this doesn't apply, does it?

docs/src/guidelines.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+## Expected behaviour
+
+The following is a non-exhaustive list of expected behaviour for methods.
+
+- All in place methods should also return the modified vectors.
+- Vector inputs should have the correct size. If necessary, pass them using views or slices.
```
dpo: the user should pass them

docs/src/guidelines.md
abelsiqueira: I meant the minimum information that should be set in case you're not using a simple unconstrained model. I'm gonna explain it better.

docs/src/guidelines.md
abelsiqueira: Between lines 8 and 9?

docs/src/guidelines.md
abelsiqueira: I'm not sure it's worth the trouble.

docs/src/guidelines.md
abelsiqueira: No, but you can still write an `ADNLPModel` equivalent to your AL and check the consistency (although some functions may need to be excluded).

dpo: I see what you mean. That'd be a bit more advanced though. Perhaps it could appear in a separate section?!

abelsiqueira: Sure. I'm creating a different example for this.

docs/src/guidelines.md
abelsiqueira: I noticed that I also need some additional counter functions, such as `increment!` and `sum_counters`, so I've written the macro to check how beneficial it is. Let me know what you think.

abelsiqueira:
Rebased

src/nlp_utils.jl
```diff
@@ -61,3 +62,32 @@ function coo_sym_prod!(rows :: AbstractVector{<: Integer}, cols :: AbstractVecto
   end
   return Av
 end
+
+"""
+    @default_inner_counters Model inner
```
dpo: Just `@default_counters` would do it I think.

docs/src/guidelines.md
dpo: I'm not convinced this part is really relevant. It's somewhat advanced.

abelsiqueira: Do you mind if I move the whole section to the bottom and call it `Advanced testing`?

dpo: Sounds good

dpo:
Obrigado.

PR #254, opened 2020-03-01
Version 0.11.2

abelsiqueira:


abelsiqueira:
These errors on Julia 1.1 are random but frequent (since `jprod` with structure). I have no idea what's causing them. They may fail because multiple-precision tests are too slow. I've opened #255 to check that.

abelsiqueira:
Closed in favor of 0.12.0

PR #256, opened 2020-03-01
Drop Julia 1.1 from CI

abelsiqueira:


dpo:
We could drop them everywhere but then we should also update the Project.toml.

abelsiqueira:
Sure.

abelsiqueira:
I also updated version to 0.12.0 so it can be release right after this.

amontoison:
Maybe we should drop julia 1.2 too? Julia 1.4 will be available soon. (version rc2 currently)
Besides Artifacts (.jll) and the 5 arguments mul! that we want to use with Dominique for LinearOperators.jl / Krylov.jl in the future is only available for Julia >= 1.3.

abelsiqueira:
Now I dropped Julia up to 1.2 and added 1.4. I left version 0.12.0 to after #257.

PR #257, opened 2020-03-02
Add jth_hess_residual implementation

amontoison:
Is it possible to merge this before the release 0.12 of NLPModel? I will need it for `NLPModelJuMP`.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -556,7 +556,7 @@ function hprod!(nlp::AbstractNLPModel, rows::AbstractVector{<: Integer}, cols::A
 end
 
 """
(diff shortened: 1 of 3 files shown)


abelsiqueira:
Could you add a consistency test?

amontoison:
Yes, but is it not already include in `ǹls_consistency` ? https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/40a356a0d952b5636e5c015b01e21c7c376c3132/test/nls_consistency.jl#L118-L127

abelsiqueira:
I see, I thought this was the first implementation of `jth`, but it's actually just a default.

PR #258, opened 2020-03-20
Add NLP problem LINCON

amontoison:
Implementation of the model used in `NLPModelsJuMP`.

### DIFF
--- a/test/consistency.jl
+++ b/test/consistency.jl
@@ -413,7 +413,7 @@ function consistent_nlps(nlps; rtol=1.0e-8)
   # If there are inequalities, test the SlackModels of each of these models
   if nlps[1].meta.ncon > length(nlps[1].meta.jfix)
     slack_nlps = [SlackModel(nlp) for nlp in nlps]
-    consistent_functions(slack_nlps)
+    consistent_functions(slack_nlps, exclude=[cons])
     consistent_counters(slack_nlps)
     @printf("✓")
   else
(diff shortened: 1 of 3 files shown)


test/consistency.jl
```diff
@@ -413,7 +413,7 @@ function consistent_nlps(nlps; rtol=1.0e-8)
   # If there are inequalities, test the SlackModels of each of these models
   if nlps[1].meta.ncon > length(nlps[1].meta.jfix)
     slack_nlps = [SlackModel(nlp) for nlp in nlps]
-    consistent_functions(slack_nlps)
+    consistent_functions(slack_nlps, exclude=[cons])
```
amontoison: With a `SlackModel`, lower bounds and upper bounds of constraints become lower / upper bounds of slack variables. So if both models are different constraints bounds, I cannot compare `cons` functions.

With linear constraints and a `MathOptNLPModel`, `l <= Ax + b <= u` is internally stored as `l - b <= Ax <= u - b`.

test/problems/lincon.jl
```diff
@@ -0,0 +1,152 @@
...
+
+function NLPModels.grad!(nlp :: LINCON, x :: AbstractVector, gx :: AbstractVector)
+  increment!(nlp, :neval_grad)
+  gx .= [4 * x[i]^3 for i =1:nlp.meta.nvar]
+  return gx
+end
+
+function NLPModels.hess(nlp :: LINCON, x :: AbstractVector{T}; obj_weight=one(T)) where T
+  increment!(nlp, :neval_hess)
+  return T.(diagm([12.0 * x[i]^2 for i=1:nlp.meta.nvar])) * obj_weight
```
abelsiqueira: I think you can change `12.0` to `12` and drop `T`

test/consistency.jl
abelsiqueira: Can you change the example here in `NLPModels` to `l - b <= Ax <= u - b`?

abelsiqueira:
Thank you

PR #280, opened 2020-04-20
Remove NotImplemented and use MethodError instead

abelsiqueira:
Following this idea: https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html
I also noticed an unnecessary computation of Hessians in `ADNLPModel`, I guess from the time we computed the sparsity of the Hessian there.

### DIFF
--- a/src/autodiff_model.jl
+++ b/src/autodiff_model.jl
@@ -12,7 +12,7 @@ inital estimate to the Lagrangian multipliers can also be provided.
 
 ````
 ADNLPModel(f, x0; lvar = [-∞,…,-∞], uvar = [∞,…,∞], y0 = zeros,
-  c = nothing, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
+  c = x->[], lcon = [], ucon = [], name = "Generic")
 ````
 
   - `f` - The objective function ``f``. Should be callable;
@@ -51,26 +51,23 @@ end
 
 show_header(io :: IO, nlp :: ADNLPModel) = println(io, "ADNLPModel - Model with automatic differentiation")
 
-function ADNLPModel(f, x0::AbstractVector; y0::AbstractVector = eltype(x0)[],
-                    lvar::AbstractVector = eltype(x0)[], uvar::AbstractVector = eltype(x0)[],
+function ADNLPModel(f, x0::AbstractVector;
+                    lvar::AbstractVector = fill(-Inf, length(x0)), uvar::AbstractVector = fill(Inf, length(x0)),
                     lcon::AbstractVector = eltype(x0)[], ucon::AbstractVector = eltype(x0)[],
-                    c = nothing,
+                    c = x -> eltype(x0)[], y0::AbstractVector = eltype(x0)[],
                     name::String = "Generic", lin::AbstractVector{<: Integer}=Int[])
 
+  T = eltype(x0)
   nvar = length(x0)
-  length(lvar) == 0 && (lvar = -Inf*ones(nvar))
-  length(uvar) == 0 && (uvar =  Inf*ones(nvar))
   ncon = maximum([length(lcon); length(ucon); length(y0)])
+  length(lcon) == 0 && (lcon = fill(-T(Inf), ncon))
+  length(ucon) == 0 && (ucon = fill(T(Inf), ncon))
+  length(y0) == 0 && (y0 = zeros(T, ncon))
+  @lencheck ncon lcon ucon y0
 
   nnzh = nvar * (nvar + 1) / 2
-  nnzj = 0
+  nnzj = nvar * ncon
 
-  if ncon > 0
-    length(lcon) == 0 && (lcon = -Inf*ones(ncon))
-    length(ucon) == 0 && (ucon =  Inf*ones(ncon))
-    length(y0) == 0   && (y0 = zeros(ncon))
-    nnzj = nvar * ncon
-  end
   nln = setdiff(1:ncon, lin)
 
   meta = NLPModelMeta(nvar, x0=x0, lvar=lvar, uvar=uvar, ncon=ncon, y0=y0,
--- a/src/autodiff_model.jl
+++ b/src/autodiff_model.jl
@@ -12,7 +12,7 @@ inital estimate to the Lagrangian multipliers can also be provided.
 
 ````
 ADNLPModel(f, x0; lvar = [-∞,…,-∞], uvar = [∞,…,∞], y0 = zeros,
-  c = NotImplemented, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
+  c = nothing, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
 ````
 
   - `f` - The objective function ``f``. Should be callable;
@@ -54,7 +54,7 @@ show_header(io :: IO, nlp :: ADNLPModel) = println(io, "ADNLPModel - Model with 
 function ADNLPModel(f, x0::AbstractVector; y0::AbstractVector = eltype(x0)[],
                     lvar::AbstractVector = eltype(x0)[], uvar::AbstractVector = eltype(x0)[],
                     lcon::AbstractVector = eltype(x0)[], ucon::AbstractVector = eltype(x0)[],
-                    c = (args...)->throw(NotImplementedError("cons")),
+                    c = nothing,
                     name::String = "Generic", lin::AbstractVector{<: Integer}=Int[])
 
   nvar = length(x0)
--- a/src/autodiff_nlsmodel.jl
+++ b/src/autodiff_nlsmodel.jl
@@ -7,7 +7,7 @@ compute the derivatives.
 
 ````
 ADNLSModel(F, x0, m; lvar = [-∞,…,-∞], uvar = [∞,…,∞], y0 = zeros,
-  c = NotImplemented, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
+  c = nothing, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
 ````
 
   - `F` - The residual function \$F\$. Should be callable;
@@ -33,7 +33,7 @@ function ADNLSModel(F, x0 :: AbstractVector, m :: Int;
                     name :: String = "Generic",
                     lvar :: AbstractVector = fill(-eltype(x0)(Inf), length(x0)),
                     uvar :: AbstractVector = fill( eltype(x0)(Inf), length(x0)),
-                    c = (args...)->throw(NotImplementedError("cons")),
+                    c = nothing,
                     lcon :: AbstractVector = eltype(x0)[],
                     ucon :: AbstractVector = eltype(x0)[],
                     y0 :: AbstractVector = zeros(eltype(x0), max(length(lcon), length(ucon))),
--- a/src/autodiff_nlsmodel.jl
+++ b/src/autodiff_nlsmodel.jl
@@ -7,7 +7,7 @@ compute the derivatives.
 
 ````
 ADNLSModel(F, x0, m; lvar = [-∞,…,-∞], uvar = [∞,…,∞], y0 = zeros,
-  c = nothing, lcon = [-∞,…,-∞], ucon = [∞,…,∞], name = "Generic")
+  c = x->[], lcon = [], ucon = [], name = "Generic")
 ````
 
   - `F` - The residual function \$F\$. Should be callable;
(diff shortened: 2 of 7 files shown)


src/autodiff_model.jl
```diff
@@ -52,18 +52,14 @@ end
 function ADNLPModel(f, x0::AbstractVector; y0::AbstractVector = eltype(x0)[],
                     lvar::AbstractVector = eltype(x0)[], uvar::AbstractVector = eltype(x0)[],
                     lcon::AbstractVector = eltype(x0)[], ucon::AbstractVector = eltype(x0)[],
-                    c = (args...)->throw(NotImplementedError("cons")),
+                    c = nothing,
```
dpo, line 55: How is this going to work if the user calls `cons()`?

abelsiqueira: `ERROR: MethodError: objects of type Nothing are not callable`

dpo: I'm not really convinced by this. I found the NotImplementedError helpful. In the blog, the author makes an error (passing a vector instead of a matrix) and implemented an error message that isn't helpful. I get the gist but in this particular case, does `ERROR: MethodError: objects of type Nothing are not callable` really help understand what's wrong?

abelsiqueira: Maybe for this specific model, having `NotImplemented` may be helpful, but the overall change is for the API. `NotImplemented` would say `foo not implemented` and the current way says something like `no method matching obj(::DummyModel, ::Array{Float64,1})` and it shows the closest candidates.

dpo: Could we use something like `c = () -> nothing` so as to trigger a MethodError? Or `c = function cons end`?

abelsiqueira: Sure, I'll think of something.

abelsiqueira: I think the most descriptive way is to create two models `ADNLPUncModel` and `ADNLPConModel`, both `<: ADNLPModels <: AbstractNLPModel`, and then change the functions to receive either `ADNLPModels` or `ADNLPConModel`.

vepiteski: Why not consider that unconstrained models simply provide 0 as constraints? Of course then one has to ensure that when not empty,  lcon <= 0 <= ucon. Perhaps two constructors achieve that.
Imagine benchmarking a solver for constrained problems, and including some unconstrained instances in the testset, the solver would crash because there is no constraint?

dpo: Actually that's a great point. We do define lvar, lcon, etc as empty arrays. So c should also return an empty array.

dpo: Also there's an inconsistency with the lvar/uvar here and https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/src/nlp_types.jl#L101

abelsiqueira: Thanks, that's much better. I've added a new commit changing that and defining `lvar` and `uvar` correctly on the keyword default values.

dpo: Thanks!

src/autodiff_nlsmodel.jl
```diff
@@ -31,7 +31,7 @@ function ADNLSModel(F, x0 :: AbstractVector, m :: Int;
                     name :: String = "Generic",
                     lvar :: AbstractVector = fill(-eltype(x0)(Inf), length(x0)),
                     uvar :: AbstractVector = fill( eltype(x0)(Inf), length(x0)),
-                    c = (args...)->throw(NotImplementedError("cons")),
+                    c = nothing,
```
dpo, line 34: Same here.

dpo: And y0 here is not as in ADNLPModel.

abelsiqueira:
More corrections

abelsiqueira:
:robot: Testing breakage of this pull request

| Package Name | master | stable |
|--|--|--|
| AmplNLReader | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689542) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689543) | 
| CUTEst | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689544) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689545) | 
| CaNNOLeS | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689546) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689547) | 
| NLPModelsIpopt | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689548) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689549) | 
| NLPModelsJuMP | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689550) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689551) | 
| QuadraticModels | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689552) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689553) | 
| SolverTools | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689554) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677689555) |

abelsiqueira:
:robot: Testing breakage of this pull request

| Package Name | master | stable |
|--|--|--|
| AmplNLReader | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/677749188) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122087) | 
| CUTEst | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122088) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122089) | 
| CaNNOLeS | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122090) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122091) | 
| NLPModelsIpopt | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122092) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122093) | 
| NLPModelsJuMP | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122094) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122095) | 
| QuadraticModels | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122096) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122097) | 
| SolverTools | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122098) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/679122099) |

src/autodiff_model.jl
```diff
@@ -49,29 +49,20 @@ mutable struct ADNLPModel <: AbstractNLPModel
...
-  nnzj = 0
+  nnzj = nvar * ncon
 
-  if ncon > 0
-    length(lcon) == 0 && (lcon = -Inf*ones(ncon))
-    length(ucon) == 0 && (ucon =  Inf*ones(ncon))
-    length(y0) == 0   && (y0 = zeros(ncon))
-    nnzj = nvar * ncon
+  if !(length(lcon) == length(ucon) == length(y0))
+    error("lcon, ucon and y0 need to be the same length")
```
dpo, line 65: This is causing an error if the user specifies only one of those arguments. Example: https://travis-ci.org/github/JuliaSmoothOptimizers/NLPModels.jl/jobs/677749203#L271.

That's annoying because we now must specify all of them or none of them.  Alternatively, we could specify `ncon`.

dpo: Also we should use `@lencheck`.

abelsiqueira:
:robot: Testing breakage of this pull request

| Package Name | master | stable |
|--|--|--|
| AmplNLReader | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183821) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183822) | 
| CUTEst | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183823) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183824) | 
| CaNNOLeS | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183825) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183826) | 
| NLPModelsIpopt | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183827) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183828) | 
| NLPModelsJuMP | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183829) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183830) | 
| QuadraticModels | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183831) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183832) | 
| SolverTools | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183833) | [](https://travis-ci.org/JuliaSmoothOptimizers/NLPModels.jl/jobs/690183834) |

abelsiqueira:
Finally updated. Not sure why Cirrus is failing, but FreeBSD on travis is passing.

src/autodiff_model.jl
abelsiqueira: Updated to accept one, two, or three arguments.

src/autodiff_model.jl
```diff
@@ -51,26 +51,23 @@ end
...
 show_header(io :: IO, nlp :: ADNLPModel) = println(io, "ADNLPModel - Model with automatic differentiation")
 
-function ADNLPModel(f, x0::AbstractVector; y0::AbstractVector = eltype(x0)[],
-                    lvar::AbstractVector = eltype(x0)[], uvar::AbstractVector = eltype(x0)[],
+function ADNLPModel(f, x0::AbstractVector;
+                    lvar::AbstractVector = fill(-Inf, length(x0)), uvar::AbstractVector = fill(Inf, length(x0)),
                     lcon::AbstractVector = eltype(x0)[], ucon::AbstractVector = eltype(x0)[],
-                    c = (args...)->throw(NotImplementedError("cons")),
+                    c = x -> eltype(x0)[], y0::AbstractVector = eltype(x0)[],
                     name::String = "Generic", lin::AbstractVector{<: Integer}=Int[])
```
dpo, line 58: This looks fine but I find there's a discrepancy between the way we treat the bounds (without explicit AbstractVectors) and the constraints. What if there were another ADNLPModel constructor that also takes `ncon` and `c` as positional parameters, and where `lcon`, `ucon` and `y0` have default values defined with `fill!(-Inf, ncon)`, etc.?

abelsiqueira: I'm creating constructors for each traditional situation:
- `ADNLPModel(f, x0, lvar, uvar)`
- `ADNLPModel(f, x0, ncon, ceq)`
- `ADNLPModel(f, x0, cineq, lcon, ucon)`
- `ADNLPModel(f, x0, lvar, uvar, ncon, ceq)`
- `ADNLPModel(f, x0, lvar, uvar, cineq, lcon, ucon)`
what do you think?

abelsiqueira: In a different PR

dpo: What does `ADNLPModel(f, x0, ncon, ceq)` do? Construct a zero rhs?

It may be enough to write
* `ADNLPModel(f, x0, lvar, uvar)`
* `ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon)`

no?

abelsiqueira: Yes, zero rhs. I think we can reduce to
- `ADNLPModel(f, x0, lvar, uvar)`
- `ADNLPModel(f, x0, c, lcon, ucon)`
- `ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon)`

dpo: Ok!

dpo:
Green lights!

PR #291, opened 2020-05-11
Update NLS examples with linear residuals

amontoison:
Linear residuals are placed before nonlinear residuals. It's what we already do with the constraints. We should be able to check the consistency of NLS models that make the distinction between linear and nonlinear residuals.


abelsiqueira:
Thanks

dpo:
So it didn't merge automatically despite the two approvals?

abelsiqueira:
I didn't even notice

PR #303, opened 2020-08-05
Update the documentation about NLPModelsJuMP

amontoison:


.travis.yml
```diff
@@ -83,6 +83,6 @@ jobs:
       after_success: skip
 
 after_success:
-  - julia -e 'if Sys.islinux() && string(VERSION)[1:3] == "1.4"
```
amontoison, line 86: Abel, do you have an idea for avoiding this update at each new release of Julia ?

dpo: Use just 1 ?!

.travis.yml
```diff
@@ -83,6 +83,6 @@ jobs:
...
+  - julia -e 'if Sys.islinux() && string(VERSION)[1:3] == "1.5"
```
abelsiqueira, line 86:
```suggestion
  - julia -e 'if Sys.islinux()
```
There is no harm in building every linux, so we can just avoid deciding which release.

.travis.yml
amontoison: `1` is only used by Travis. `VERSION` is a variable defined by Julia and it is 1.0.5, 1.4.2, 1.5.0, etc...

abelsiqueira:
Thanks

PR #309, opened 2020-11-12
Create ci.yml

abelsiqueira:
I removed the 32 bits part from https://github.com/jump-dev/Ipopt.jl/blob/master/.github/workflows/ci.yml


dpo:
Should we remove Windows from the GitHub action to save some minutes, since we already have appveyor?

dpo:
Frankly we could use CircleCI to build on Linux (2500 mins) and only use Actions for macOS (which is not free anywhere else)?!

dpo:
see #311

dpo (approved):

PR #311, opened 2020-11-12
fix CircleCI config

dpo:


dpo:
Where should we add the coverage check?

abelsiqueira:
> Where should we add the coverage check?

You mean which CI? It should be on Linux because it's faster/cheaper. Github actions for MacOS consumes 10x the amount of minutes that linux consumes.

dpo:
We're going to burn through our November minutes just setting up CI... :-(

abelsiqueira (approved):

PR #312, opened 2020-12-04
use GitHub Actions across the board

dpo:


abelsiqueira (approved):

dpo:
We need to allow some failures.

PR #318, opened 2021-01-10
Fix bug in `feasibility_residual`

tmigot:


tmigot:
fix bug in `hprod!` function.

abelsiqueira:
Thanks. I had to squash because of the commit history. Normally, what we do use branch and rebasing to keep the history clean.

PR #319, opened 2021-01-11
Improve hessian tests

tmigot:
Add hessian functions in test problems, so that we can also test consistency for `hprod` of nls problems.

### DIFF
--- a/test/nls_consistency.jl
+++ b/test/nls_consistency.jl
@@ -163,7 +163,7 @@ function consistent_nls_functions(nlss; rtol=1.0e-8, exclude=[])
   end
 end
 
-function consistent_nlss(nlss; exclude=[hess, hprod, hess_coord])
+function consistent_nlss(nlss; exclude=[hess, hess_coord])
   consistent_nls_counters(nlss)
   consistent_counters(nlss)
   consistent_nls_functions(nlss, exclude=exclude)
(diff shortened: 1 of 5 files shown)


tmigot:
Test failed because of PR #318

tmigot:
The diff still thinks that the modifications in `src/feasibility_residual.jl` are new, but they were already merged in PR #318.

abelsiqueira:
> The diff still thinks that the modifications in `src/feasibility_residual.jl` are new, but they were already merged in PR #318.

That's because you haven't rebased your branch on top of the updated `master`. See here: https://github.com/tmigot/NLPModels.jl/network

Looking at your graph, I recommend
- git checkout master # Go to your master branch
- git branch tmp # Create a backup of that branch
- git reset --hard jso/master # Reset your master to jso master (jso is our JSO remote)
- git rebase master imp_hess_tests # rebase

As a rule of thumb, your master should never be ahead of JSO's.

abelsiqueira:
PS. These git commands can destroy your work, backup if you're not sure.

tmigot:
> PS. These git commands can destroy your work, backup if you're not sure.

Thanks for your help! It looks alright now.

abelsiqueira (approved):
I was a little worried about the addition of `hess_structure` and `_coord` functions for NLS because we are not implementing it and we are not expecting them to be implemented. I'm not sure if adding these functions is a bad thing, though, so I am approving and leaving the decision of whether to keep them as well to @dpo.

dpo:
We’ve had requests for hess, hess_structure & co. for LLS problems. It’s a bad idea to use them but I guess we could allow the user to implement and test them.

dpo:
Thank you

PR #320, opened 2021-03-03
Add ghjvprod implementation and tests

tmigot:
- add API doc
- doc in NLPModels.jl
- autodiff implementations
- feasibility/slack/lls implementations
- implement in test set
- consistency test


src/NLPModels.jl
```diff
@@ -427,12 +427,20 @@ end
 
 function jth_hprod! end
 
+"""
+   gHv = ghjvprod(nlp, x, g, v)
+Return the vector whose i-th component is g' ∇²cᵢ(x) v.
```
dpo, line 432: Don’t we need a blank line before the `Return` so the markdown is formatted correctly?

src/autodiff_model.jl
```diff
@@ -241,3 +241,12 @@ function hprod!(nlp :: ADNLPModel, x :: AbstractVector, y :: AbstractVector, v :
   Hv .= ForwardDiff.derivative(t -> ForwardDiff.gradient(ℓ, x + t * v), 0)
   return Hv
 end
+
+function ghjvprod!(nlp :: ADNLPModel, x :: AbstractVector, g :: AbstractVector, v :: AbstractVector, gHv :: AbstractVector) 
+  @lencheck nlp.meta.nvar x g v
+  @lencheck nlp.meta.ncon gHv
+  increment!(nlp, :neval_hprod)
+  ℓ(x) = nlp.c(x)
```
dpo, line 249: Necessary?

dpo (approved):
Thanks!

src/NLPModels.jl
tmigot: You are right, thanks !

src/autodiff_model.jl
tmigot: You are right, thanks !

PR #323, opened 2021-03-09
Use NLPModelsCore and NLPModelsTestUtils

abelsiqueira:
Creating the PR to run the tests

cf. @dpo @tmigot


abelsiqueira:
Github actions (and appveyor) don't trigger when there is a conflict, so I will add `ghjvprod`

abelsiqueira:
Following a different strategy

PR #324, opened 2021-03-10
GitHub Actions Breakage tests

abelsiqueira:
~Trying to fix breakage tests on GitHub actions. May take a while until it's working.~ Went better than expected.

Closes #313

PR #325, opened 2021-03-10
Add exclude to consistent_nlps and defaults to ghjvprod

abelsiqueira:
~Waiting for #324, do not merge.~ Ready.

ghjvprod breaks a few other packages, so it is initially excluded from the API consistency tests.


abelsiqueira:
I realized that I should add `exclude=[]` to some calls now, but I started something else on NLPModels.jl, so I'll get back to this soon after

PR #347, opened 2021-04-14
Showcasing the blue style

abelsiqueira:
@dpo, this is the blue style, committed manually.


src/nlp/counters.jl
```diff
@@ -48,39 +48,39 @@ end
 
 Increment counter `s` of problem `nlp`.
 """
-function increment!(nlp :: AbstractNLPModel, s :: Symbol)
-  setproperty!(nlp.counters, s, getproperty(nlp.counters, s) + 1)
+function increment!(nlp::AbstractNLPModel, s::Symbol)
+  return setproperty!(nlp.counters, s, getproperty(nlp.counters, s) + 1)
```
abelsiqueira, line 52: One main difference is always using return

src/nlp/meta.jl
```diff
@@ -63,92 +63,93 @@ The following keyword arguments are accepted:
...
-                        lnet=Int[],
-                        nlin=length(lin),
-                        nnln=length(nln),
-                        nnnet=length(nnet),
-                        nlnet=length(lnet),
-                        minimize=true,
-                        nlo=1,
-                        islp=false,
-                        name="Generic")
+  nvar::Int
```
abelsiqueira, line 66: It doesn't like some empty lines

test/nls/api.jl
```diff
@@ -21,8 +20,12 @@
   @test hess_residual(nls, x, w) ≈ HF(x, w)
   @test jprod_residual(nls, x, v) ≈ JF(x) * v
   @test jtprod_residual(nls, x, w) ≈ JF(x)' * w
-  @test jprod_residual!(nls, jac_structure_residual(nls)..., jac_coord_residual(nls, x), v, Jv) ≈ JF(x) * v
-  @test jtprod_residual!(nls, jac_structure_residual(nls)..., jac_coord_residual(nls, x), w, Jtw) ≈ JF(x)' * w
+  @test jprod_residual!(
+    nls, jac_structure_residual(nls)..., jac_coord_residual(nls, x), v, Jv
```
abelsiqueira, line 24: And then there is space, it prefers to leave it in one line.

test/nls/simple-model.jl
```diff
@@ -19,71 +19,95 @@ x₀ = ones(n).
...
 end
 
 function SimpleNLSModel()
-  meta = NLPModelMeta(2, x0=ones(2), name="Simple NLS Model", lvar=zeros(2), uvar=ones(2), ncon=3, lcon=[0.0; 0.0; 1.0], ucon=[Inf; Inf; 1.0], nnzj=6)
-  nls_meta = NLSMeta(2, 2, nnzj=3, nnzh=1)
+  meta = NLPModelMeta(
+    2;
+    x0=ones(2),
+    name="Simple NLS Model",
+    lvar=zeros(2),
```
abelsiqueira, line 32: The blue style doesn't like spaces around `=`, but the default does.

abelsiqueira (commented):
Also worth noting that this is not the complete blue style, only what's implemented so far.

dpo:
Let’s know what you prefer. I’m happy as long as we can keep 2 spaces indentation. I tend to prefer long lines to fixed line length, but it’s not a deal breaker for me. Consistency is more important.

PR #351, opened 2021-05-21
Use new LinearOperators

abelsiqueira:
Has to be updated after https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/pull/180 is released and tagged.


amontoison:
@abelsiqueira, I updated your PR with the last modifications of LinearOperators.jl.

dpo (approved):

PR #353, opened 2021-06-01
Parametrize structures

geoffroyleconte:


geoffroyleconte:
I think depending on the installed julia version, vectors are displayed `Vector{Float64}` or `Array{Float64,1}`, which leads to some failures for the show tests

src/nlp/meta.jl
```diff
@@ -161,25 +162,27 @@ struct NLPModelMeta <: AbstractNLPModelMeta
     @lencheck nnnet nnet
     @lencheck nlnet lnet
     @rangecheck 1 ncon lin nln nnet lnet
+    S = typeof(x0)
+    @assert S == typeof(lvar) == typeof(uvar) == typeof(y0) == typeof(lcon) == typeof(ucon)
```
abelsiqueira, line 166: It would be better to set `x::S`, `lvar::S`, etc., and use `T = eltype(S)`, but I understood it's not possible to define the starting values for that case.

I wonder if we should define this constructor as `NLPModelaMeta{T, S}` and define a constructor 
```
NLPModelMeta(n, x0::S; kwargs...) = NLPModelMeta{eltype(S), S}(n, x0=x0; kwargs...)
```
This would certainly break more, but it would be clearer. @dpo, what do you think?

abelsiqueira (changes requested):
Thanks. I made some comments.

src/nlp/meta.jl
```diff
@@ -62,11 +62,11 @@ The following keyword arguments are accepted:
 - `islp`: true if the problem is a linear program
 - `name`: problem name
 """
-struct NLPModelMeta <: AbstractNLPModelMeta
+struct NLPModelMeta{T,S} <: AbstractNLPModelMeta{T,S}
```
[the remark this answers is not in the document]
geoffroyleconte: I think it's OK since our algorithms will not be based on updates of these vectors, but I'm not 100% sure. @amontoison can you confirm this?

src/nlp/meta.jl
```diff
@@ -117,10 +117,10 @@ struct NLPModelMeta <: AbstractNLPModelMeta
   name::String
 
   function NLPModelMeta(
-    nvar;
-    x0 = zeros(nvar),
-    lvar = -Inf * ones(nvar),
-    uvar = Inf * ones(nvar),
+    nvar::Int;
+    x0::AbstractVector{T} = zeros(nvar),
```
tmigot, line 121: I think it should be `zeros(T, nvar)` and similar for `lvar` and `uvar` here.

src/nlp/meta.jl
```diff
@@ -131,9 +131,9 @@ struct NLPModelMeta <: AbstractNLPModelMeta
     nlvoi = 0,
     nwv = 0,
     ncon = 0,
-    y0 = zeros(ncon),
-    lcon = -Inf * ones(ncon),
-    ucon = Inf * ones(ncon),
+    y0::AbstractVector{T} = zeros(ncon),
```
tmigot, line 134: Same as before.

src/nls/meta.jl
```diff
@@ -33,15 +33,16 @@ end
 function NLSMeta(
   nequ::Int,
   nvar::Int;
-  x0::AbstractVector = zeros(nvar),
+  x0::AbstractVector{T} = zeros(nvar),
```
tmigot, line 36: `zeros(T, nvar)`

test/nlp/utils.jl
```diff
@@ -19,7 +19,7 @@ end
 
 @testset "Increase coverage of default_NLPcounters" begin
   @default_counters SuperNLPModel model
-  nlp = SuperNLPModel(SimpleNLPModel())
+  nlp = SuperNLPModel{Float64, Vector{Float64}}(SimpleNLPModel())
```
tmigot, line 22: It might be useful to also have a test involving another type than the default `Float64`.

tmigot (commented):
Great PR, thanks ! I just have a couple of comments.

src/nlp/meta.jl
abelsiqueira: There is no inference of type `T` because `x0` is a keyword argument. It would need to be `NLPModelMeta{T}`, or either `T` or `x0` would need to be a mandatory argument. That's why I suggested `NLPModelMeta{T,S}` and a constructor with `x0` as mandatory above. What do you think?

src/nlp/meta.jl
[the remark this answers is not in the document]
amontoison: If you don't update some vectors or you want to use scalar indexing with them you should store them on CPU.

src/nlp/meta.jl
tmigot: Ohhh, ok sorry for that.

Completely agree with you with the `x0` as a mandatory argument.

geoffroyleconte:
Thank you for your reviews. What should I do for the show tests with julia 1.3?

test/nlp/show.jl
```diff
@@ -3,7 +3,7 @@
   io = IOBuffer()
   show(io, nlp)
   showed = String(take!(io))
-  expected = """  SimpleNLPModel
+  expected = """  SimpleNLPModel{Float64, Vector{Float64}}
```
abelsiqueira, line 6:
```suggestion
  storage_type = VERSION < v"1.6" ? "Vector{Float64}" : "Array{Float64, 1}"
  expected = """  SimpleNLPModel{Float64, $storage_type}
```

abelsiqueira: My suggestion for the `@show` fix.

test/nlp/simple-model.jl
```diff
@@ -32,6 +32,22 @@ function SimpleNLPModel()
   return SimpleNLPModel(meta, Counters())
 end
 
+function SimpleNLPModel32()
```
abelsiqueira, line 35: You can define the constructor above as
```
function SimpleNLPModel(::Type{T}) where T
```
and another constructor `SimpleNLPModel() = SimpleNLPModel(Float64)`.

abelsiqueira (approved):
Thanks for the update. I suggested how to fix the `@show` issue.

test/nlp/show.jl
```diff
@@ -3,7 +3,8 @@
   io = IOBuffer()
   show(io, nlp)
   showed = String(take!(io))
-  expected = """  SimpleNLPModel
+  storage_type = typeof(nlp)
```
geoffroyleconte, line 6: I directly took the type of nlp, because it looks like nightly uses spaces, I though it would be simpler like this

abelsiqueira: I think that's fine.

abelsiqueira:
It would be good to create some PRs fixing some packages so we know if this needs some adjustments. Probably NLPModelsModifiers, NLPModelsTest and JSOSolvers. I hope to have some time tomorrow.

abelsiqueira:
@dpo looks like the bot is not working exactly as planned. It was supposed to merge only after both of us approved it. Since `revert` creates an ugly commit history, let me know if you want to revert manually or if we should move ahead with a release.

PR #356, opened 2021-06-04
Update Version

geoffroyleconte:


dpo:
This only breaks ... everything. Let's stay optimistic!

dpo (approved):

geoffroyleconte:
Should I register this commit or wait?

dpo:
I think you can go ahead.

tmigot:
Yes, so we can go ahead in fixing everything :).

PR #357, opened 2021-06-17
Update breakage list

abelsiqueira:


tmigot (commented):
Thank you for adding the new/missing packages.
It is also the opportunity to rename DCI:
"JuliaSmoothOptimizers/DCI.jl" is now "JuliaSmoothOptimizers/DCISolver.jl"

abelsiqueira:
Good catch

abelsiqueira:
I'm saving this PR to when we have all (most?) other packages fixed, to achieve maximum serotonin.

PR #358, opened 2021-06-21
Update default arguments for NLPModelMeta

geoffroyleconte:
With this PR and after updating the compatibilities of NLPModelsModifiers, SolverCore and QuadraticModels, I am able to create a QuadraticModel with ` CuArray{Float64, 1}` vectors.


geoffroyleconte:
It looks like the breakage tests fail because of a compatibility issue with versions 0.15 and 0.16

abelsiqueira:
Looks great, but we should try to fix the other packages to NLPModels 0.16 to see if this really is breaking or not.

dpo (approved):

PR #371, opened 2021-07-20
Add getter for `NLPModelMeta` and `NLSMeta`

tmigot:
This way calling `get_minimize(meta)` would no longer allocate the 2nd call, while `meta.minimize` would always.
For each attribute `foo` in the meta, there is a `get_foo` function.

Note that there is a priority for NLS over NLP for copies in NLPMeta and NLSMeta, e.g. `get_x0(nls)` will look for `nls.nls_meta.x0`.


dpo (approved):
Great! Thanks!

dpo:
Hope it was ok to merge!

tmigot:
> Hope it was ok to merge!

Yes, it was ok to merge. The `SolverBenchmark.jl` failure comes from a different issue following the upgrade in `BenchmarkProfiles.jl`.

PR #372, opened 2021-07-23
Simplify meta

tmigot:
Remove everything related to integer variables, network constraints, and multiple objectives in the meta

PR #373, opened 2021-07-26
Version 0.17.2

tmigot:
Following #372 to fix `AmplNLReader.jl`.


abelsiqueira:
I just realized the README needs to be updated as well, can you do that too?

tmigot:
> I just realized the README needs to be updated as well, can you do that too?

You are right, thanks!

abelsiqueira:
Thanks

Issue #377, opened 2021-09-13
Better support for linear constraints


abelsiqueira:
As discussed with @tmigot and @dpo, create separate API for linear and nonlinear constraints.

tmigot:
As a small note, in Knitro (see [PR 72 in NLPModelsKnitro.jl](https://github.com/JuliaSmoothOptimizers/NLPModelsKnitro.jl/pull/72)), the constraint is expected to have the form:
```
x -> c(x) + A * x
```
where `A` is given separately.

PR #379, opened 2021-10-06
API for linear constraints

tmigot:
#377 
Add new entries in the meta:
- `lin_nnzj` and `nln_nnzj`, respectively, the number of nonzeros in the Jacobian matrix of linear/nonlinear constraints.

Add new functions in the API:
- `lin_cons` and `nln_cons`
- and then propagate this for all the Jacobian functions.

There is no default implementation of these functions, i.e. we do not use the fact that by default there are no linear constraints.


tmigot:
This is ready to review @abelsiqueira @dpo .
As discussed, there is now a small issue with `jtprod`, where I don't really see a way to avoid allocating. So, I added a [comment L.516](https://github.com/tmigot/NLPModels.jl/blob/97092b2bb8ab05068880422c0efa0cd06b3d70d5/src/nlp/api.jl#L516)

src/nlp/api.jl
```diff
@@ -128,15 +176,70 @@ end
...
+    nln_jac_structure!(nlp, view(rows, nln_ind), view(cols, nln_ind))
+    view(rows, nln_ind) .+= nlp.meta.nlin
+  end
+  return rows, cols
+end
+
+"""
+    (rows,cols) = lin_jac_structure(nlp)
+
+Return the structure of the linear constraint's Jacobian in sparse coordinate format.
```
dpo, line 194:
```suggestion
Return the structure of the linear constraints Jacobian in sparse coordinate format.
```

src/nlp/api.jl
```diff
@@ -128,15 +176,70 @@ end
...
+function lin_jac_structure(nlp::AbstractNLPModel)
+  rows = Vector{Int}(undef, nlp.meta.lin_nnzj)
+  cols = Vector{Int}(undef, nlp.meta.lin_nnzj)
+  lin_jac_structure!(nlp, rows, cols)
+end
+
+"""
+    lin_jac_structure!(nlp, rows, cols)
+
+Return the structure of the linear constraint's Jacobian in sparse coordinate format in place.
```
dpo, line 205:
```suggestion
Return the structure of the linear constraints Jacobian in sparse coordinate format in place.
```

src/nlp/api.jl
```diff
@@ -128,15 +176,70 @@ end
...
+"""
+function lin_jac_structure! end
+
+"""
+    (rows,cols) = nln_jac_structure(nlp)
+
+Return the structure of the nonlinear constraint's Jacobian in sparse coordinate format.
```
dpo, line 212:
```suggestion
Return the structure of the nonlinear constraints Jacobian in sparse coordinate format.
```

src/nlp/api.jl
```diff
@@ -128,15 +176,70 @@ end
...
+function nln_jac_structure(nlp::AbstractNLPModel)
+  rows = Vector{Int}(undef, nlp.meta.nln_nnzj)
+  cols = Vector{Int}(undef, nlp.meta.nln_nnzj)
+  nln_jac_structure!(nlp, rows, cols)
+end
+
+"""
+    nln_jac_structure!(nlp, rows, cols)
+
+Return the structure of the nonlinear constraint's Jacobian in sparse coordinate format in place.
```
dpo, line 223:
```suggestion
Return the structure of the nonlinear constraints Jacobian in sparse coordinate format in place.
```

src/nlp/api.jl
```diff
@@ -128,15 +176,70 @@ end
...
+"""
+function nln_jac_structure! end
 
 """
     vals = jac_coord!(nlp, x, vals)
 
 Evaluate ``J(x)``, the constraint's Jacobian at `x` in sparse coordinate format,
```
dpo, line 230:
```suggestion
Evaluate ``J(x)``, the constraints Jacobian at `x` in sparse coordinate format,
```

src/nlp/api.jl
```diff
@@ -161,6 +264,68 @@ function jac(nlp::AbstractNLPModel, x::AbstractVector)
   sparse(rows, cols, vals, nlp.meta.ncon, nlp.meta.nvar)
 end
 
+"""
+    vals = lin_jac_coord!(nlp, x, vals)
+
+Evaluate ``J(x)``, the linear constraint's Jacobian at `x` in sparse coordinate format,
```
dpo, line 270:
```suggestion
Evaluate ``J(x)``, the linear constraints Jacobian at `x` in sparse coordinate format,
```

src/nlp/api.jl
```diff
@@ -161,6 +264,68 @@ function jac(nlp::AbstractNLPModel, x::AbstractVector)
...
+rewriting `vals`.
```
dpo, line 271:
```suggestion
overwriting `vals`.
```

src/nlp/api.jl
```diff
@@ -161,6 +264,68 @@ function jac(nlp::AbstractNLPModel, x::AbstractVector)
...
+"""
+function lin_jac_coord! end
+
+"""
+    vals = lin_jac_coord(nlp, x)
+
+Evaluate ``J(x)``, the linear constraint's Jacobian at `x` in sparse coordinate format.
```
dpo, line 278: etc, etc, there are more of the same

src/nlp/api.jl
```diff
@@ -215,6 +387,114 @@ jprod!(
...
+
+"""
+    Jv = lin_jprod!(nlp, x, v, Jv)
+
+Evaluate ``J(x)v``, the linear Jacobian-vector product at `x` in place.
+"""
+function lin_jprod! end
+
+"""
+    Jv = lin_jprod!(nlp, rows, cols, vals, v, Jv)
```
dpo, line 409: What is the use of this variant?

src/nlp/api.jl
```diff
@@ -215,6 +387,114 @@ jprod!(
...
+)
+  @lencheck nlp.meta.lin_nnzj rows cols vals
+  @lencheck nlp.meta.nvar v
+  @lencheck nlp.meta.nlin Jv
+  increment!(nlp, :neval_jprod)
+  coo_prod!(rows, cols, vals, v, Jv)
+end
+
+"""
+    Jv = lin_jprod!(nlp, x, rows, cols, v, Jv)
```
dpo, line 430: How do you compute a product without the values?

src/nlp/api.jl
```diff
@@ -392,6 +799,250 @@ function jac_op!(
...
+
+Return the linear Jacobian at `x` as a linear operator.
+The resulting object may be used as if it were a matrix, e.g., `J * v` or
+`J' * v`.
+"""
+function lin_jac_op(nlp::AbstractNLPModel{T, S}, x::AbstractVector{T}) where {T, S}
+  @lencheck nlp.meta.nvar x
+  Jv = S(undef, nlp.meta.nlin)
+  Jtv = S(undef, nlp.meta.nvar)
+  return jac_op!(nlp, x, Jv, Jtv)
```
dpo, line 813:
```suggestion
  return lin_jac_op!(nlp, x, Jv, Jtv)
```

src/nlp/api.jl
```diff
@@ -392,6 +799,250 @@ function jac_op!(
...
+
+Return the nonlinear Jacobian at `x` as a linear operator.
+The resulting object may be used as if it were a matrix, e.g., `J * v` or
+`J' * v`.
+"""
+function nln_jac_op(nlp::AbstractNLPModel{T, S}, x::AbstractVector{T}) where {T, S}
+  @lencheck nlp.meta.nvar x
+  Jv = S(undef, nlp.meta.nnln)
+  Jtv = S(undef, nlp.meta.nvar)
+  return jac_op!(nlp, x, Jv, Jtv)
```
dpo, line 935:
```suggestion
  return nln_jac_op!(nlp, x, Jv, Jtv)
```

dpo (changes requested):
First round...

abelsiqueira:
More of a higher-level comment: should we be appending `lin` instead of pretending?
Does it make a difference anywhere? Docs?

src/nlp/api.jl
tmigot: I also changed the others in the same file.

src/nlp/api.jl
tmigot: Thanks

src/nlp/api.jl
tmigot: It is for cases where one needs to compute (or one has computed) the jacobian first to compute the product.

The same exists with `jprod!(nlp, rows, cols, vals, v, Jv)` and `nln_jprod!(nlp, rows, cols, vals, v, Jv)`.

src/nlp/api.jl
tmigot: Not entirely sure for this one... I guess the implementation of the function should compute it directly making some use of the structure.

The same exists with `jprod!(nlp, rows, cols, v, Jv)` and `nln_jprod!(nlp, rows, cols, v, Jv)`.

tmigot: After discussion, this will be removed by #382

PR #381, opened 2021-11-08
Change default `nln` in `NLPModelMeta`

tmigot:
Ensure compatibility between `lin` and `nlin`.

### DIFF
--- a/src/nlp/meta.jl
+++ b/src/nlp/meta.jl
@@ -116,9 +116,6 @@ struct NLPModelMeta{T, S} <: AbstractNLPModelMeta{T, S}
     nln_nnzj = nvar * ncon,
     nnzh = nvar * (nvar + 1) / 2,
     lin = Int[],
-    nln = 1:ncon,
-    nlin = length(lin),
-    nnln = length(nln),
     minimize = true,
     islp = false,
     name = "Generic",
@@ -129,9 +126,7 @@ struct NLPModelMeta{T, S} <: AbstractNLPModelMeta{T, S}
 
     @lencheck nvar x0 lvar uvar
     @lencheck ncon y0 lcon ucon
-    @lencheck nlin lin
-    @lencheck nnln nln
-    @rangecheck 1 ncon lin nln
+    @rangecheck 1 ncon lin
     # T = eltype(x0)
 
     ifix = findall(lvar .== uvar)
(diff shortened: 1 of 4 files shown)


src/nlp/meta.jl
```diff
@@ -122,7 +122,7 @@ struct NLPModelMeta{T, S} <: AbstractNLPModelMeta{T, S}
     end
 
     @lencheck nvar x0 lvar uvar
-    @lencheck ncon y0 lcon ucon
+    @lencheck ncon y0 lcon ucon union(lin, nln)
```
abelsiqueira, line 125: Discussed: Remove `nln`, `nlin` and `nnln` from input.

tmigot:
Good news: it simplifies building the Meta and ensure more fields are correct
Bad news: everything is breaking...

abelsiqueira:
Wow, that's one of the best breaks so far.
@dpo, opinions?

tmigot:
We can also merge this after/with #379 , which will break things.

dpo:
Wow. Pretty thorough breakage! The changes look great! That's how it should be.

tmigot:
@abelsiqueira I also updated the meta in the `NLSMeta` since we have to review many packages.
Actually, the breakage is not that large, since some packages are breaking just because they depend on `NLPModesModifiers` and `ADNLPModels`.

The following are ready:			
NLPModelsModifiers		https://github.com/JuliaSmoothOptimizers/NLPModelsModifiers.jl#fix-meta					
ADNLPModels		https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl#fix-meta					
LLSModels		https://github.com/JuliaSmoothOptimizers/LLSModels.jl#fix-meta					
CUTEst		https://github.com/JuliaSmoothOptimizers/CUTEst.jl#fix-meta					
NLPModelsJuMP		https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl#fix-meta									
QuadraticModels		https://github.com/JuliaSmoothOptimizers/QuadraticModels.jl#fix-meta					

We can up `NLPModels` to 0.18 after merging this PR. I will create issues along the way suggesting to implement the linear API.

abelsiqueira (approved):
Thanks, this is great.
I think you mention setting`nln` in the guidelines. If so, can you open an issue so we remember to fix it, please?

PR #382, opened 2021-11-15
Remove matrix-product with the structure only

tmigot:
I let:
```
jac_op!(nlp, x, rows, cols, Jv, Jtv)
hess_op!(nlp, x, rows, cols, Hv; obj_weight=1.0)
hess_op!(nlp, x, y, rows, cols, Hv; obj_weight=1.0)
```
and
```
jac_op_residual!(nls, x, rows, cols, Jv, Jtv)
```
in case initializing LinearOperators benefit from the structure.

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -201,21 +201,6 @@ function jprod!(
 end
 
 """
-    Jv = jprod!(nlp, x, rows, cols, v, Jv)
-
-Evaluate ``J(x)v``, the Jacobian-vector product at `x` in place.
-`(rows, cols)` should be the Jacobian structure in triplet format.
-"""
-jprod!(
-  nlp::AbstractNLPModel,
-  x::AbstractVector,
-  ::AbstractVector{<:Integer},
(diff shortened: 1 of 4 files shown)


tmigot:
@dpo @abelsiqueira it's trickier than expected to know if any of this is useful because it is tested in `NLPModelsTest.jl`. So, all the packages using these tests fail.

abelsiqueira:
But at least we know that Percival and DCISolver don't need it.

abelsiqueira:
We can update NLPModelsTest removing the test before this.

dpo:
> in case initializing LinearOperators benefit from the structure.

I don't think it's the case. IMO, those should be removed too.

tmigot:
I prepared the two PR to fix the breakage of `AmplNLReader.jl` and `NLPModelsModifiers.jl`, so this can be reviewed.

dpo (approved):
lgtm

abelsiqueira:
Thank you!

PR #390, opened 2021-12-20
Add counters for lin/nln functions

tmigot:


tmigot:
@abelsiqueira I believe this could close #377, what do you think ?

abelsiqueira:
Two things:
- I think that NLPModelsTest will verify the counters (I may have removed this test already), so different implementations (e.g., not having linear constraints), will break the tests. Probably nothing that can be done at this end, but if the NLPModelsTest break, that's a good explanation.
- Should we enforce `neval_cons = neval_cons_lin + neval_cons_nln`? Should we remove `neval_cons`? `sum_counters` will sum everything, so it may affect comparisons.

tmigot:
> Two things:
> 
> * I think that NLPModelsTest will verify the counters (I may have removed this test already), so different implementations (e.g., not having linear constraints), will break the tests. Probably nothing that can be done at this end, but if the NLPModelsTest break, that's a good explanation.

Apparently, nothing will be breaking (apart from a `show` test in NLPModelModifiers.jl).

> * Should we enforce `neval_cons = neval_cons_lin + neval_cons_nln`? Should we remove `neval_cons`? 

I don't really like the idea of forcing this, this could be a test though.
Removing `neval_cons` would enforce the fact that one needs to use the _lin and _nln functions.
Maybe we can deprecate the functions impacted,  `neval_cons`, `neval_jac`, `neval_jprod` and `neval_jtprod`?

> `sum_counters` will sum everything, so it may affect comparisons.
If we remove `neval_cons` etc..., this is ok. Otherwise, we could make different variants of `sum_counters`.

abelsiqueira:
The NLPModelsTest part the could produce errors is here: https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/blob/main/src/nlp/consistency.jl#L76
The problem would be: I implement a model and compare against HS14 from NLPModelsTest. Assume that NLPModelsTest is updated to implement the lin and nln parts separately. If my new model does not uses linear and nonlinear constraints, that consistency check will fail. So we should update NLPModelsTest to be consistent by either ignoring the tests, or forcing equality, etc.

About the second point, I'm not sure what to do. Maybe I'm overreacting and we should dismiss it, but we should probably decide some rules or guidelines at least and document them. For instance, comparing a solver that uses the separate functions vs another that doesn't. How will they behave on models that (i) implement lin and nln; (ii) implement only nln; (iii) implement only the full version. Should we change the sum_counters behavior?

test/nlp/show.jl
```diff
@@ -6,22 +6,25 @@
...
+   All variables: ████████████████████ 2      All constraints: ████████████████████ 2     
+            free: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0                 free: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0     
+           lower: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0                lower: ██████████⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 1     
+           upper: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0                upper: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0     
+         low/upp: ████████████████████ 2              low/upp: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0     
+           fixed: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0                fixed: ██████████⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 1     
+          infeas: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0               infeas: ⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 0     
+            nnzh: ( 33.33% sparsity)   2               linear: ██████████⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 1     
+                                                    nonlinear: ██████████⋅⋅⋅⋅⋅⋅⋅⋅⋅⋅ 1     
+                                                         nnzj: (  0.00% sparsity)   4
```
dpo, line 18: are these trailing spaces expected?

tmigot:
> However, `sum_couters` should not count things twice.

Would it be ok if `sum_counters` were summing up all counters without `lin` or `nln` suffix?

The only alternative (apart from the sum of everything as is now) I see is to satisfy `neval_cons = sum(neval_cons_lin .&& neval_cons_nln)` in the function `sum_counters`, but it is not accurate whenever somebody re-implements `cons!` as well.

abelsiqueira:
> Would it be ok if sum_counters were summing up all counters without lin or nln suffix?

I think this is the solution.

dpo:
> Would it be ok if sum_counters were summing up all counters without lin or nln suffix?

I think it should sum all counters *except* those without lin or nln suffix. If I only call con_lin and con_nln, then, neval_cons will be zero, which is not representative of what I evaluated.

dpo (approved):

PR #393, opened 2022-02-10
inline increment! to avoid allocations

dpo:
I found while benchmarking that `increment!()` allocates 2 bytes each time. It seems inlining solves the issue.


abelsiqueira:
Great find.

PR #394, opened 2022-02-15
hotfix gpu

amontoison:
`findall` returns an error with an empty `CuVector`.
I was unable to create a NLPModelMeta without this modification.


abelsiqueira:
Thanks

PR #396, opened 2022-04-19
Use abstract type for meta API

tmigot:
https://github.com/JuliaSmoothOptimizers/NLPModelsModifiers.jl/issues/51


abelsiqueira (approved):
Thanks

PR #401, opened 2022-05-03
Fix linear API 2

tmigot:
Another bug. The shift in the rows was only valid when linear constraints are given first.

It now works with: 
- https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/tree/add-linear-API
- https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/tree/add-linear-api
- https://github.com/JuliaSmoothOptimizers/LLSModels.jl/tree/add-linearAPI
- https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/tree/add-linearAPI

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -185,13 +185,13 @@ function jac_structure!(
   lin_ind = 1:(nlp.meta.lin_nnzj)
   nlp.meta.nlin > 0 && jac_lin_structure!(nlp, view(rows, lin_ind), view(cols, lin_ind))
   for i in lin_ind
-    rows[i] += count(<(nlp.meta.lin[rows[i]]), nlp.meta.nln)
+    rows[i] += count(x < nlp.meta.lin[rows[i]] for x in nlp.meta.nln)
   end
   if nlp.meta.nnln > 0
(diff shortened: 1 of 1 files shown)


src/nlp/api.jl
```diff
@@ -184,10 +184,15 @@ function jac_structure!(
   @lencheck nlp.meta.nnzj rows cols
   lin_ind = 1:(nlp.meta.lin_nnzj)
   nlp.meta.nlin > 0 && jac_lin_structure!(nlp, view(rows, lin_ind), view(cols, lin_ind))
+  for i in lin_ind
+    rows[i] += count(<(nlp.meta.lin[rows[i]]), nlp.meta.nln)
```
abelsiqueira, line 188: I felt this was too obscure and I made a small benchmark and apparently the following is faster and allocates less (?!). But mostly, I think it is clearer
```suggestion
    rows[i] += count(x < nlp.meta.lin[rows[i]] for x in nlp.meta.nln)
```

abelsiqueira (approved):
Except for my comment, feel free to merge. Thanks

PR #407, opened 2022-06-03
bump version to 0.19

dpo:


tmigot:
Should we also add versions of package here https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/docs/Project.toml to avoid the doc from failing with a breaking version ?

dpo:
Ok. Which version(s) of ADNLPModels do you suggest?

tmigot:
0.3 looks good. We will update later with the 0.4 release.

tmigot (approved):

Issue #409, opened 2022-06-06
Fix docs


abelsiqueira:
Need `ADNLPModels 0.3.X` to be released with a compatible version, or updated after `ADNLPModels 0.4`.

PR #411, opened 2022-08-03
use strict=true in docs

dpo:


tmigot:
This fails because of #409 as ADNLPModels has not reached up-to-date NLPModels release.

tmigot:
So, the error comes from duplicate documentation.

The `autodocs` block in the references page print all the docstrings. However, in pages like:
https://juliasmoothoptimizers.github.io/NLPModels.jl/dev/models/
https://juliasmoothoptimizers.github.io/NLPModels.jl/dev/tools/
https://juliasmoothoptimizers.github.io/NLPModels.jl/dev/api/
https://juliasmoothoptimizers.github.io/NLPModels.jl/dev/internals/
we also call some docstring.

I see two solutions:
- we remove the docstrings from this page.
- we replace the docstrings by links to the reference page.

@dpo @abelsiqueira Any idea or opinion?

dpo:
> we replace the docstrings by links to the reference page.

tmigot:
https://juliasmoothoptimizers.github.io/NLPModels.jl/previews/PR411/

PR #412, opened 2022-08-10
Up doc environment

tmigot:


tmigot:
@dpo @abelsiqueira I re-ran the breakage tests to have an overview of the update to NLPModels 0.18/0.19, and we are now up-to-date !! Only the linear API in AmplNLReader is missing.

abelsiqueira:
That's great news. Thanks for all the hard work.

Issue #385, opened 2021-12-13
Separate the default implementations from the api file


abelsiqueira:
Leave only the definition and docs in the `api.jl` file and create a new file with the default implementations. We can duplicate the function "declaration".

PR #417, opened 2022-08-31
Add in-place `obj` and `grad!` for nls

tmigot:


src/nls/api.jl
```diff
@@ -388,25 +388,61 @@ function hess_op_residual!(nls::AbstractNLSModel, x::AbstractVector, i::Int, Hiv
...
-function objgrad!(nls::AbstractNLSModel, x::AbstractVector, g::AbstractVector)
+function grad!(nls::AbstractNLSModel{T, S}, x::AbstractVector, g::AbstractVector) where {T, S}
+  @lencheck nls.meta.nvar x g
+  increment!(nls, :neval_grad)
+  Fx = S(undef, nls.nls_meta.nequ)
+  return grad!(nls, x, g, Fx)
+end
+
+"""
+    f, g = grad!(nls, x, g, Fx)
```
amontoison, line 429:
```suggestion
    f, g = objgrad!(nls, x, g, Fx)
```

src/nls/api.jl
```diff
@@ -388,25 +388,61 @@ function hess_op_residual!(nls::AbstractNLSModel, x::AbstractVector, i::Int, Hiv
   )
 end
 
-function obj(nls::AbstractNLSModel, x::AbstractVector)
+"""
+    f = obj(nls::AbstractNLSModel, x::AbstractVector, Fx::AbstractVector)
```
amontoison, line 392:
```suggestion
    f = obj(nls::AbstractNLSModel, x::AbstractVector)
    f = obj(nls::AbstractNLSModel, x::AbstractVector, Fx::AbstractVector)
```

src/nls/api.jl
```diff
@@ -388,25 +388,61 @@ function hess_op_residual!(nls::AbstractNLSModel, x::AbstractVector, i::Int, Hiv
...
+
+Evaluate `f(x)`, the objective function of `nls::AbstractNLSModel`. `Fx` is overwritten with the value of the residual `F(x)`.
+"""
```
amontoison, line 395:
```suggestion
"""
function obj end

```

src/nls/api.jl
```diff
@@ -388,25 +388,61 @@ function hess_op_residual!(nls::AbstractNLSModel, x::AbstractVector, i::Int, Hiv
...
 
-function grad!(nls::AbstractNLSModel, x::AbstractVector, g::AbstractVector)
+function obj(nls::AbstractNLSModel{T, S}, x::AbstractVector) where {T, S}
+  @lencheck nls.meta.nvar x
+  Fx = S(undef, nls.nls_meta.nequ)
+  return obj(nls, x, Fx)
+end
+
+"""
+    g = grad!(nls, x, g, Fx)
```
amontoison, line 410:
```suggestion
    g = grad!(nls, x, g)
    g = grad!(nls, x, g, Fx)
```

amontoison (changes requested):
Maybe a vector `Fx` should be allocated directly when we create a NLS model and stored in the model.

dpo:
I agree that ideally, `Fx` would be preallocated inside the model. As we know, the downside is that that doesn't play nicely with automatic differentiation.

src/nls/api.jl
tmigot: Does it change something to have the empty function?

amontoison: If you only have one docstring for a function with multiple methods, it recommended to add the docstring above the empty function.
In the REPL, it will display the docstring for the `obj` function without specific argument. 
```
julia> ?
help> obj
```

amontoison: Otherwise, you should add a docstring for all methods.

amontoison (approved):

amontoison:
I forgot that you have the bot for this package -_-

src/nls/api.jl
tmigot: I see; actually, there is an issue connected to this https://github.com/JuliaSmoothOptimizers/NLPModels.jl/issues/385

amontoison: See QRMumps.jl for example:
https://github.com/JuliaSmoothOptimizers/QRMumps.jl/blob/main/src/QRMumps.jl

PR #423, opened 2022-09-23
[WIP] Create specific operator for hprod

abelsiqueira:
First jab at avoiding the implicit change of `hess_op` is up for discussion.

To use it on a code like Trunk, instead of writing `H = hess_op!(nlp, x, Hv)`, we will just write `H.x .= x`. We can probably wrap that under `update!(H, x)`, but it does not seem necessary.

Issues:

- Verbose
- There is currently no way to pass an existing `x` to create an `HprodOperator`, but that could be changed, if we want to. The API is becoming too convoluted, though.
- The `hess_op!` with `rows, cols, vals` will still suffer from the same implicit issue, unless we also create a `HprodOperator` for it.

### DIFF
--- a/src/NLPModels.jl
+++ b/src/NLPModels.jl
@@ -37,6 +37,7 @@ Base type for a nonlinear least-squares model.
 """
 abstract type AbstractNLSModel{T, S} <: AbstractNLPModel{T, S} end
 
+include("nlp/operator.jl")
 for f in ["utils", "api", "counters", "meta", "show", "tools"]
   include("nlp/$f.jl")
   include("nls/$f.jl")
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -1225,16 +1225,7 @@ function hess_op!(
   obj_weight::Real = one(eltype(x)),
 )
   @lencheck nlp.meta.nvar x Hv
-  prod! = @closure (res, v, α, β) -> begin
-    hprod!(nlp, x, v, Hv; obj_weight = obj_weight)
-    if β == 0
-      @. res = α * Hv
-    else
-      @. res = α * Hv + β * res
-    end
-    return res
-  end
-  return LinearOperator{eltype(x)}(nlp.meta.nvar, nlp.meta.nvar, true, true, prod!, prod!, prod!)
+  return HprodOperator!(nlp, copy(x), Hv, obj_weight=obj_weight)
 end
 
 """
--- a/src/nlp/operator.jl
+++ b/src/nlp/operator.jl
@@ -1,3 +1,5 @@
+export ModelOperator, update!
+
 using FastClosures, LinearOperators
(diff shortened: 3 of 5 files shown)


tmigot:
I don't like the `H.x .= x` since for the following type of operator that wouldn't do anything:
```
function hess_op!(
  nlp::GridapPDENLPModel,
  x::AbstractVector,
  Hv::AbstractVector;
  obj_weight::Real = one(eltype(x)),
)
  @lencheck nlp.meta.nvar x Hv
  rows = nlp.pdemeta.Hrows
  cols = nlp.pdemeta.Hcols
  vals = hess_coord!(nlp, x, nlp.workspace.Hvals, obj_weight = obj_weight)
  return hess_op!(nlp, rows, cols, vals, Hv)
end
```

dpo:
This is looking good, thanks!

> The hess_op! with rows, cols, vals will still suffer from the same implicit issue, unless we also create a HprodOperator for it.

Could it be another constructor?

tmigot:
Ok, so if I understand. Whenever we reimplemented `hess_op!`, we would now have to specialize `HprodOperator!(nlp, x, Hv)` and `update!(::ModelOperator, x)`. Is that right?

If we want to reuse the structure for another `nlp` of the same size, we would also need a `update!(::ModelOperator, nlp)` and a way to broadcast `nlp` assuming they respect some conditions.

abelsiqueira:
Some comments on the comments

> I don't like the H.x .= x since for the following type of operator that wouldn't do anything:
> ...
> Ok, so if I understand. Whenever we reimplemented hess_op!, we would now have to specialize `HprodOperator!(nlp, x, Hv)` and `update!(::ModelOperator, x)`. Is that right?

Yes, I think it will be necessary, but only models that feel the need to update `hess_op!` itself. They would, instead, update `HprodOperator`, so the additional work is the same (well, almost the same).

> If we want to reuse the structure for another `nlp` of the same size, we would also need a `update!(::ModelOperator, nlp)` and a way to broadcast `nlp` assuming they respect some conditions.

Maybe we need a `update!(::Solver, nlp)`. Consider the following:

```julia
solver = TrunkSolver(nlp) # solver.H exists
nlp = Problem1()
output = solve!(solver, nlp)
# do things
nlp = Problem2() 
### Option 1: Update solver.H ???
### Option 2: Update solver
output = solve!(solver, nlp)
```

Option 1 requires specific knowledge of the solver. Option 2 can be a no-op for many solvers. Also, we can accept keyword arguments to update other solver aspects, like subsolver.

@dpo
> > The hess_op! with rows, cols, vals will still suffer from the same implicit issue, unless we also create a HprodOperator for it.
> Could it be another constructor?

Not like this, because `ModelOperator` stores `x`. We would need to store `rows, cols, vals`, I think.
Do we have any use cases for this already? I think our solvers using `rows, cols, vals` are all factorization.

PR #424, opened 2022-09-27
put NLPModelMeta constructor outside of struct

paraynaud:
@dpo, I continue my road to fully integrate PartitionedVectors in PartiallySeparableNLPModels.
For that, I need to define a NLPMeta relying on PartitionedVectors.
The current constructor doesn't support PartitionedVector, even if I provide directly some PartitionedVectors x0, uvar, xvar..., the following code:
```julia
ifix = findall(lvar .== uvar)
ilow = findall((lvar .> T(-Inf)) .& (uvar .== T(Inf)))
iupp = findall((lvar .== T(-Inf)) .& (uvar .< T(Inf)))
irng = findall((lvar .> T(-Inf)) .& (uvar .< T(Inf)) .& (lvar .< uvar))
ifree = findall((lvar .== T(-Inf)) .& (uvar .== T(Inf)))
iinf = findall(lvar .> uvar)
```
cause errors.

This PR puts the NLPMeta constructor outside the struct, which allows to define a NLPMeta dedicated to PartitionedVectors in PartiallySeparableNLPModels.jl.

### DIFF
--- a/src/nlp/meta.jl
+++ b/src/nlp/meta.jl
@@ -143,7 +143,6 @@ function NLPModelMeta{T, S}(
   @lencheck nvar x0 lvar uvar
   @lencheck ncon y0 lcon ucon
   @rangecheck 1 ncon lin
-  # T = eltype(x0)
 
   ifix = findall(lvar .== uvar)
   ilow = findall((lvar .> T(-Inf)) .& (uvar .== T(Inf)))
--- a/src/nlp/meta.jl
+++ b/src/nlp/meta.jl
@@ -112,106 +112,106 @@ struct NLPModelMeta{T, S} <: AbstractNLPModelMeta{T, S}
   minimize::Bool
   islp::Bool
   name::String
+end
 
-  function NLPModelMeta{T, S}(
-    nvar::Int;
-    x0::S = fill!(S(undef, nvar), zero(T)),
-    lvar::S = fill!(S(undef, nvar), T(-Inf)),
-    uvar::S = fill!(S(undef, nvar), T(Inf)),
-    nlvb = nvar,
-    nlvo = nvar,
-    nlvc = nvar,
-    ncon = 0,
-    y0::S = fill!(S(undef, ncon), zero(T)),
-    lcon::S = fill!(S(undef, ncon), T(-Inf)),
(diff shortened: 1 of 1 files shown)


src/nlp/meta.jl
```diff
@@ -112,106 +112,106 @@ struct NLPModelMeta{T, S} <: AbstractNLPModelMeta{T, S}
...
-      lin,
-      nln,
-      minimize,
-      islp,
-      name,
-    )
+  @lencheck nvar x0 lvar uvar
+  @lencheck ncon y0 lcon ucon
+  @rangecheck 1 ncon lin
+  # T = eltype(x0)
```
tmigot, line 146:
```suggestion
```

tmigot:
Thanks @paraynaud ! This looks like a good idea. I have a suggestion that would improve further changes.
If only the following block is a problem:
```
  ifix = findall(lvar .== uvar)
  ilow = findall((lvar .> T(-Inf)) .& (uvar .== T(Inf)))
  iupp = findall((lvar .== T(-Inf)) .& (uvar .< T(Inf)))
  irng = findall((lvar .> T(-Inf)) .& (uvar .< T(Inf)) .& (lvar .< uvar))
  ifree = findall((lvar .== T(-Inf)) .& (uvar .== T(Inf)))
  iinf = findall(lvar .> uvar)
```
Could we put this in a separate function? (Actually, we would improve the code as it is reused for constraints)
Then, you would just have to make specialize this function instead of making a new constructor.
My point is that if we modify NLPModelMeta later on, we wouldn't have to modify your version as well.

What do you think?

paraynaud:
```quote
 My point is that if we modify NLPModelMeta later on, we wouldn't have to modify your version as well.
```
You're right.

```quote
Could we put this in a separate function? 
```
I think we can.
It should be parametrized by 2 AbstractVectors.
I will define `ifix, ilow, iupp, irng, ifree, iinf` differently when it is called on two PartitionedVectors.
I don't know how you want to proceed, but the `else` part of contraints works fine.

tmigot (approved):

dpo (approved):
Thanks @paraynaud !

I wonder if it'd be more efficient to make a single pass over lvar/uvar and lcon/ucon and fill in the indexes ourselves instead of calling findall many times.

PR #426, opened 2022-11-14
v0.19.2

paraynaud:
@dpo @tmigot , I didn't see any breaking changes in the commits since the latest release, so I updated the version to 0.19.2.


tmigot (approved):
It looks good, thanks @paraynaud

tmigot:
Should we remove the auto-merge here as well @dpo ?

PR #434, opened 2023-04-17
remove non-existent variable from docstring

tpapp:
trivial fix

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -460,7 +460,7 @@ function jprod_nln!(
 end
 
 """
-    Jtv = jtprod(nlp, x, v, Jtv)
+    Jtv = jtprod(nlp, x, v)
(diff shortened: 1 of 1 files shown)


tmigot (commented):
Nice catch, thanks! 

Could you also fix the same for `jtprod_lin` and `jtprod_nln` in the same PR?

https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/c2108b8b16d3c0e7a240a5e1155b1f5129623672/src/nlp/api.jl#L520

https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/c2108b8b16d3c0e7a240a5e1155b1f5129623672/src/nlp/api.jl#L560

tmigot (approved):
Thanks @tpapp

PR #438, opened 2023-04-28
Update documentation reset!

### LINKED ISSUE #437: reset! function
Hi, I don't know if this is an issue, but when using `reset!(nlp)`, the data of `nlp` are also reset (not mentionned in the doc).
Should `reset!` not call `reset_data!`? Or should the doc of `reset!`be modified?

https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/0b36c4f59ebf4d61f73203e9917549256c999661/src/nlp/counters.jl#L114

geoffroyleconte:
Fixes #437

### DIFF
--- a/src/nlp/counters.jl
+++ b/src/nlp/counters.jl
@@ -109,7 +109,7 @@ end
 """
     reset!(nlp)
 
-Reset evaluation count in `nlp`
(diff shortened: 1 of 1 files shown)


tmigot (approved):
Thanks !

PR #441, opened 2023-06-09
docs: enable linkcheck

dpo:


tmigot (approved):
Thanks! We should probably have it in all packages.

PR #450, opened 2023-09-23
Update CI scripts

amontoison:


tmigot:
Thanks @amontoison for the suggestion. I think @abelsiqueira is actually working on something there so he should be able to help.

abelsiqueira:
I am, indeed, this: https://github.com/abelsiqueira/COPIERTemplate.jl/
Here is an example of how it will look like: https://github.com/JuliaSmoothOptimizers/JSOSuite.jl/pull/126
But I do not object to merging this now.

amontoison:
Thanks Tangi and Abel, I merge the PR and next time we will use [COPIERTemplate.jl](https://github.com/abelsiqueira/COPIERTemplate.jl/).

PR #451, opened 2024-01-13
Update models documentation

tmigot:


tmigot:
We should maybe consider replacing this by a link to the jso.dev website to avoid being out-of-sync.

Issue #384, opened 2021-12-13
Improve implementation of `jtprod`


tmigot:
Following #379 , the implementation of `jtprod` [L. 475](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/289b81ceea0961cb60a6e536d32311114b09405b/src/nlp/api.jl#L475) allocate by default when the problem has linear and nonlinear constraints.

The optimal way is to re-implement this method when creating a `NLPModel`, but we could think of alternatives.
For instance, having a 5-arguments `mul!` with `jtprod`.

PR #453, opened 2024-02-26
Avoid views in jtprod!

tmigot:
This would solve https://github.com/JuliaSmoothOptimizers/NLPModelsModifiers.jl/issues/77

```
using NLPModels, NLPModelsModifiers, NLPModelsTest
nlp = NLSLC()
nls = FeasibilityFormNLS(nlp)
x = get_x0(nls)
y = get_y0(nls)
Jtv = similar(x)
jtprod!(nls, x, y, Jtv)
@allocated jtprod!(nls, x, y, Jtv) # 0
```

Connected to https://github.com/JuliaSmoothOptimizers/NLPModels.jl/issues/384 .


tmigot:
The breakage fails because of the new releases of NLPModelsModifiers and to be NLPModelsTest ; I will re-run this when they arrive.

dpo (approved):

Issue #454, opened 2024-03-26
Uniform returned type in the NLPModel API


tmigot:
I think we should clarify the returned type in the API. 

Essentially, when calling `grad(nlp::AbstractNLPModel{T, S}, x::V) where {T, S, V}` we have two options:
- i) The returned type is S
- ii) The returned type is V

In general, I think assuming S == V is too restrictive.
Personally, I prefer option ii) as it is easier to implement multi-precision.

@amontoison @abelsiqueira @dpo What do you think? i) or ii) or ... ?

## Illustration

There are some incoherences right now, for instance, the functions `grad(nlp, x)` or `residual(nls,x)` do not have the same behavior.

```
using CUDA, NLPModels, NLPModelsTest # https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl#use-gpu

nls = LLS(CuArray{Float32})
x32 = nls.meta.x0
CUDA.allowscalar() do 
  residual(nls, x32)
end
x64 = fill!(CuArray{Float64}(undef, 2), 0)
CUDA.allowscalar() do 
  residual(nls, x64) # ERROR: MethodError: no method matching residual(::LLS{Float32, CuArray{Float32}}, ::CuArray{Float64, 1, CUDA.Mem.DeviceBuffer})
end

CUDA.allowscalar() do 
  grad(nls, x64) # Return a CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}
end
```

This is connected to https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/pull/105

amontoison:
- I think that i) is the best solution. The gradient should have the same type of the model.

tmigot:
Oh sorry, I fixed my suggestion. I liked ii) better as it allows to have a model and still evaluate in another precision.

Is there a particular reason @amontoison ?

dpo:
I feel it should really be
```julia
grad(nlp::AbstractNLPModel{T, S}, x::S) where {T, S}  # -> return a value of type S
```
and there should be conversion methods if the type of `x` is not `S`.

tmigot:
The views are a good point. The current implementation of `grad` does this:
```
function grad(nlp::AbstractNLPModel, x::AbstractVector)
  @lencheck nlp.meta.nvar x
  g = similar(x)
  return grad!(nlp, x, g)
end
```
to keep the type of `x` and avoid the potential issue with `views`. This is actually a third option, iii).

[...]

PR #455, opened 2024-03-26
Remove `Vector` implicit assumption

tmigot:
In some places, we assume that the vectors are of type `Vector{T}` which is not necessary the case in practice.

Connected to #454

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -101,9 +101,9 @@ function cons_nln! end
 
 function jth_con end
 
-function jth_congrad(nlp::AbstractNLPModel, x::AbstractVector, j::Integer)
+function jth_congrad(nlp::AbstractNLPModel, x::S, j::Integer) where {S}
   @lencheck nlp.meta.nvar x
-  g = Vector{eltype(x)}(undef, nlp.meta.nvar)
+  g = S(undef, nlp.meta.nvar)
   return jth_congrad!(nlp, x, j, g)
 end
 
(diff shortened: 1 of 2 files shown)


tmigot:
Breakage:
- [ ] https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/tree/avoid-sim fixes ADNLPModels
- [ ] https://github.com/JuliaSmoothOptimizers/SolverTest.jl/issues/64

amontoison (approved):

abelsiqueira (approved):
Thanks!

src/nlp/api.jl
```diff
@@ -116,10 +116,10 @@ function jth_sparse_congrad end
 
 Evaluate ``f(x)`` and ``c(x)`` at `x`.
 """
-function objcons(nlp, x)
+function objcons(nlp::AbstractNLPModel{T, S}, x::AbstractVector) where {T, S}
   @lencheck nlp.meta.nvar x
   f = obj(nlp, x)
-  c = nlp.meta.ncon > 0 ? cons(nlp, x) : eltype(x)[]
+  c = nlp.meta.ncon > 0 ? cons(nlp, x) : S(undef, 0)
```
dpo, line 122:
```suggestion
  c = cons(nlp, x)
```
`cons()` already does the right thing here.

dpo (changes requested):
Much cleaner, thank you! Just one minor comment, and the PR can be merged.

PR #462, opened 2024-04-30
Documentation tweaks (fixes #460)

### LINKED ISSUE #460: Documentation: inconsistency in whether constraints are zero
Consider [HS62](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/65bda82047287f09d261b86c968256eb21ae8df0/src/PureJuMP/hs62.jl), for which the constraint is
```julia
x[1] + x[2] + x[3] - 1 == 0
```
and the initial value is `x0 = [0.7, 0.2, 0.1]` which sums to 1. Thus you expect the initial constraint value to be zero. However:
```julia
julia> using NLPModels, NLPModelsJuMP, OptimizationProblems.PureJuMP

julia> nlp = MathOptNLPModel(hs62());

julia> cons(nlp, nlp.meta.x0)
1-element Vector{Float64}:
 0.9999999999999999
```

My suspicion is that the bug is in this package, but I'm unsure. It seems likely that it's neglecting to incorporate the constant offset in the JuMP model?

timholy:
This eliminates the inconsistency in `c_i = 0` vs `c_L <= c_i <= c_U` noted in #460.
It also:
- adds Manifest.toml to `.gitignore`
- standardizes on `y` rather than `λ` for Lagrange multipliers
- clarifies or polishes wording in various places
- adds `NLPModels` to `docs/Project.toml` (this is standard and might
  allow simplification of your `workflow`s)

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -11,16 +11,16 @@ The general form of the optimization problem is
 ```math
 \begin{aligned}
 \min \quad & f(x) \\
-& c_i(x) = c_{V_i}, \quad i \in E, \\
-& c_{L_i} \leq c_i(x) \leq c_{U_i}, \quad i \in I, \\
+& c_i(x) = c_{E_i}, \quad i \in {\cal E}, \\
+& c_{L_i} \leq c_i(x) \leq c_{U_i}, \quad i \in {\cal I}, \\
 & \ell \leq x \leq u,
 \end{aligned}
 ```
 where ``f:\mathbb{R}^n\rightarrow\mathbb{R}``,
 ``c:\mathbb{R}^n\rightarrow\mathbb{R}^m``,
-``E\cup I = \{1,2,\dots,m\}``, ``E\cap I = \emptyset``,
+``{\cal E}\cup {\cal I} = \{1,2,\dots,m\}``, ``{\cal E}\cap {\cal I} = \emptyset``,
 and
-``c_{V_i}, c_{L_i}, c_{U_i}, \ell_j, u_j \in \mathbb{R}\cup\{\pm\infty\}``
+``c_{E_i}, c_{L_i}, c_{U_i}, \ell_j, u_j \in \mathbb{R}\cup\{\pm\infty\}``
 for ``i = 1,\dots,m`` and ``j = 1,\dots,n``.
 
 For computational reasons, we write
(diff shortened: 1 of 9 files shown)


docs/src/index.md
```diff
@@ -11,7 +11,7 @@ The general form of the optimization problem is
 ```math
 \begin{aligned}
 \min \quad & f(x) \\
-& c_i(x) = 0, \quad i \in E, \\
+& c_i(x) = c_{V_i}, \quad i \in E, \\
```
timholy, line 14: I'm unsure this is the best solution, feel free to suggest alternatives. I considered jumping straight to the general statement, but it is so common to separate out equalities and inequalities that I do see value in the way things are stated now.

timholy:
Bump

tmigot:
Sorry @timholy for the delay, thanks a lot for the contribution. This is really appreciated. I will review it this weekend.

docs/src/models.md
```diff
@@ -19,7 +19,7 @@ If you want to create your own interface, check these [Guidelines](@ref).
 - [NLPModelsJuMP.jl](https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl):
   For problems modeled using [JuMP.jl](https://github.com/jump-dev/JuMP.jl).
 - [QuadraticModels.jl](https://github.com/JuliaSmoothOptimizers/QuadraticModels.jl):
-  For problems with quadratic and linear structure.
+  For problems with quadratic and linear structure (LCQP).
```
tmigot, line 22:
```suggestion
  For problems with linear constraints and a quadratic objective (LCQP).
```

docs/src/index.md
tmigot: I like your suggestion. Does the `V` stand for something? Otherwise, I would suggest `c_{E_i}` for Equality, which would match `i \in E`.

tmigot (changes requested):
Thanks a lot @timholy for the careful review of the documentation. I just had two minor comments.

timholy:
There's a lot I like about switching from `V` to `E`. But I guess one reason against `E_i` is that

$$E_i \quad \forall i \in E$$

seems slightly weird: `E` is being used both as an index set and as a constraint-value set. (The `V` is "value".)

timholy:
What about using ${\cal E}$ and ${\cal I}$ for the index sets?

tmigot:
> e index sets?

That's a great idea, and very close to what is done in optimization literature.

tmigot (approved):
LGTM, thanks again @timholy for the PR

PR #463, opened 2024-05-24
Call cons! in objcons! for unconstrained

tmigot:
This unifies the behavior with `objcons`. Otherwise, we have the strange behavior of having `neval_cons(nlp) = 0` after calling objcons!.


dpo:
This seems completely illogical to me. Of course I expect `neval_cons = 0` on an unconstrained problem. An algorithm for unconstrained optimization that attempts to evaluate the constraints has a serious existential problem.

tmigot:
This was unifying the behavior of the API, I make a new issue to discuss this.

PR #467, opened 2024-06-22
`objcons` call `objcons!`

tmigot:
Similar to objgrad


dpo (commented):
Seems like there’s a compatibility issue: https://github.com/JuliaSmoothOptimizers/NLPModels.jl/actions/runs/9626307588/job/26552129044#step:6:154

tmigot:
> Seems like there’s a compatibility issue: https://github.com/JuliaSmoothOptimizers/NLPModels.jl/actions/runs/9626307588/job/26552129044#step:6:154

It is just AmplNLReader since the linear API and CUTEst https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/327 . The other packages are now up-to-date.

Issue #469, opened 2024-07-23
Update an NLPModelMeta


MaxenceGollier:
Hello, I am trying to work on Quasi-Newton approximations of NLPModels. 
Given some `nlp::NLPModel`, I would like to construct a `SpectralGradientModel(nlp)` i.e an NLPModel where I approximate the hessian $\mathcal{H}_f \approx \sigma I$. For this, I want to update the NLPModelMeta so that `nlp.meta.nnzh = nlp.meta.nvar`, so I want to create a copy of nlp.meta but with just one field modified. Since NLPModelMeta is an immutable struct, I can't just set `nlp.meta.nnzh = nlp.meta.nvar`. Since there are quite a lot of keyword arguments and since only one field is modified, I don't want to write each time
```
new_meta = NLPModelMeta( nnzh = nlp.meta.nvar, field1 = nlp.meta.field1, field2 = nlp.meta.field2, field3 = nlp.meta.field3, ...)
```
Also, I tried using [Setfields.jl](https://jw3126.github.io/Setfield.jl/stable/intro/) but this does not seem to work. 

I am quite new to Julia so I was wondering if you had an idea on how to do this.

tmigot:
Unfortunately, I don't think there is a very clean solution here. We could add a new constructor for NLPModelMeta that simplifies this task though. Something like
```
NLPModelMeta(meta::AbstractNLPModelMeta; kwargs...) = NLPModelMeta(meta.nvar; x0 = meta.x0, etc...., name = meta.name, kwargs...)
```

MaxenceGollier:
That's what I was thinking but I wondered, if `meta` is `AbstractNLPModelMeta` but not `NLPModelMeta` for some reason, then your solution could be wrong ? I don't know if something else than `NLPModelMeta` has ever been implemented ?

tmigot:
What do you mean by it would be "wrong"?
There could be potential errors if the meta is missing some fields that are in the given keyword arguments, but anyway this would break several functions in the API.
In JSO, I am not aware of any other specialization of `AbstractNLPModelMeta`.

PR #471, opened 2024-07-25
Add new NLPModelMeta constructor

MaxenceGollier:
We discussed this in #469

Added a function allowing to copy a NLPModelMeta but with the allowance to change some fields. 
For example, if one wants to modify the number of nonzero entries of the Hessian of some nlp, (this can be the case when one wants to use a diagonal Quasi-Newton approximation of the Hessian for example), one can write
```
meta = NLPModelMeta(nlp.meta, nnzh = nlp.meta.nvar)
```

### DIFF
--- a/src/nlp/meta.jl
+++ b/src/nlp/meta.jl
@@ -214,8 +214,55 @@ function NLPModelMeta{T, S}(
   )
 end
 
-NLPModelMeta(nvar; x0::S = zeros(nvar), kwargs...) where {S} =
+NLPModelMeta(nvar::Int; x0::S = zeros(nvar), kwargs...) where {S} =
   NLPModelMeta{eltype(S), S}(nvar, x0 = x0; kwargs...)
+
+function NLPModelMeta(
+  meta::NLPModelMeta{T, S};
+  nvar::Int = meta.nvar,
+  x0::S = meta.x0,
+  lvar::S = meta.lvar,
+  uvar::S = meta.uvar,
+  nlvb = meta.nlvb,
+  nlvo = meta.nlvo,
+  nlvc = meta.nlvc,
+  ncon = meta.ncon,
+  y0::S = meta.y0,
+  lcon::S = meta.lcon,
+  ucon::S = meta.ucon,
+  nnzo = meta.nnzo,
+  nnzj = meta.nnzj,
+  lin_nnzj = meta.lin_nnzj,
+  nln_nnzj = meta.nln_nnzj,
+  nnzh = meta.nnzh,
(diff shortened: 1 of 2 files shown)


src/nlp/meta.jl
```diff
@@ -214,9 +214,56 @@ function NLPModelMeta{T, S}(
   )
 end
 
-NLPModelMeta(nvar; x0::S = zeros(nvar), kwargs...) where {S} =
+NLPModelMeta(nvar::Int; x0::S = zeros(nvar), kwargs...) where {S} =
   NLPModelMeta{eltype(S), S}(nvar, x0 = x0; kwargs...)
 
+function NLPModelMeta(
+  meta::NLPModelMeta{T, S};
```
tmigot, line 221:
```suggestion
  meta::AbstractNLPModelMeta{T, S};
```
?

src/nlp/meta.jl
```diff
@@ -214,9 +214,56 @@ function NLPModelMeta{T, S}(
...
+    ucon = ucon,
+    nnzo = nnzo,
+    nnzj = nnzj,
+    lin_nnzj = lin_nnzj,
+    nln_nnzj = nln_nnzj,
+    nnzh = nnzh,
+    lin = lin,
+    minimize = minimize,
+    islp = islp,
+    name = name
```
tmigot, line 263:
```suggestion
    name = name,
```

tmigot (changes requested):
Can you also document this new constructor here:
https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/1534231f6fd8b2e89fb9a5fe21741b82129002c0/src/nlp/meta.jl#L29
and add unit tests?

src/nlp/meta.jl
```diff
@@ -26,9 +26,11 @@ Some of their components may be infinite to indicate that the corresponding boun
 
 ---
 
-    NLPModelMeta(nvar; kwargs...)
+    NLPModelMeta(nvar::Integer; kwargs...)
+    NLPModelMeta(meta::AbstractNLPModelMeta; kwargs...)
 
 Create an `NLPModelMeta` with `nvar` variables.
+Alternatively, create an `NLPModelMeta` copy from another `AbstractNLPModelMeta` with the ability to change the fields defined in the keyword arguments.
```
tmigot, line 33:
```suggestion
Alternatively, create an `NLPModelMeta` copy from another `AbstractNLPModelMeta`.
```

tmigot (approved):
Looks good to me, thanks @MaxenceGollier

Issue #475, opened 2024-08-05
Default behavior for `nln_nnzj`


sshin23:
Hi JSO team! 

Recent change in https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/468 is causing errors in ExaModels.jl and MadNLP.jl, both of which only sets `nnzj`. I was wondering if the default behavior for `nln_nnzj` be made e.g., `nln_nnzj = nnzj - lin_nnzj`?

amontoison:
Hi Sungho!

We will have a similar issue if the user only provides `nln_nnzj`.
I'm wondering if the best solution is not fix it in MadNLP.jl and ExaModels.jl like you did.

cc @tmigot

tmigot:
Even though I agree with @amontoison , it makes sense to me that user not implementing the linear API shouldn't have to specify `nln_nnzj`. We could update version 0.21.2 with #477 if you both agree.

sshin23:
@tmigot I agree that user not implementing the linear API shouldn't have to specify `nln_nnzj`. From the user's perspective, it will be a net improvement. Of course, there's something I'm missing; we will respect NLPModels.jl dev's decision. For now, we can remove NLPModels.jl v0.21 from the compat entry of MadNLP and ExaModels

tmigot:
@sshin23 released 21.3 should fix your issue

sshin23:
Thanks, @tmigot and @amontoison

PR #476, opened 2024-08-05
Update breakage.yml

amontoison:


amontoison:
@sshin23 I added MadNLP.jl and ExaModels.jl in `breakage.yml` to check that we don't break something with minor releases.

tmigot:
Connected to #475

sshin23:
Awesome! Thank you, @amontoison  @tmigot

PR #486, opened 2024-11-18
Add an API for n-th order tensor products

amontoison:
cc @michel2323

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -1325,6 +1325,21 @@ If both `x` and `y` are provided, it computes the `n-1` projections of the **n-t
 """
 function tensor_projection end
 
+function tensor_projection(nlp::AbstractNLPModel{T, S}, n::Int, x::AbstractVector,
+                           dimension::Int, directions...) where {T, S}
+    @lencheck nlp.meta.nvar x
+    p = S(undef, nlp.meta.nvar)
+    return tensor_projection!(nlp, n, x, dimension, p, directions...)
+end
+
+function tensor_projection(nlp::AbstractNLPModel{T, S}, n::Int, x::AbstractVector,
+                           y::AbstractVector, dimension::Int, directions...) where {T, S}
+    @lencheck nlp.meta.nvar x
+    @lencheck nlp.meta.ncon y
+    p = S(undef, nlp.meta.nvar)
+    return tensor_projection!(nlp, n, x, y, dimension, p, directions...)
+end
+
 """
     tensor_projection!(nlp, n, x, dimension, p, args...)
     tensor_projection!(nlp, n, x, y, dimension, p, args...)


src/nlp/api.jl
```diff
@@ -1298,6 +1299,41 @@ function hess_op!(
...
+"""
+function tensor_projection(
+    nlp::AbstractNLPModel{T, S},
+    n::Int,
+    x::AbstractVector,
+    directions::Tuple{Int, Vararg{Int}},
+    args...
+) where {T, S}
+    @lencheck nlp.meta.nvar x
+    m = n - length(directions)
```
tmigot, line 1323: Would that make sense?
```suggestion
    directions::Tuple{Int, N},
    args...
) where {T, S, N <: Integer}
    @lencheck nlp.meta.nvar x
    m = n - N
```

src/nlp/api.jl
```diff
@@ -11,6 +11,7 @@ export jth_hess_coord, jth_hess_coord!, jth_hess
 export jth_hprod, jth_hprod!, ghjvprod, ghjvprod!
 export hess_structure!, hess_structure, hess_coord!, hess_coord
 export hess, hprod, hprod!, hess_op, hess_op!
+export tensor_projection, tensor_projection!
```
tmigot, line 14: What is your testing plan?
Update NLPModelsTest.jl would be an option, I think:
- BROWNDEN, HS13, HS5
would qualify.

src/nlp/api.jl
```diff
@@ -1298,6 +1299,41 @@ function hess_op!(
   return LinearOperator{T}(nlp.meta.nvar, nlp.meta.nvar, true, true, prod!, prod!, prod!)
 end
 
+"""
+    P = tensor_projection(nlp, n, x, directions, args...)
+
+Returns the projection of the n-th derivative of the objective of `nlp` at `x` along the specified directions.
```
tmigot, line 1305: What about `tensor_projection` ?
You should also have
`P = tensor_projection(nlp, n, x, y, directions, args...)`
as the default, where by default
`P = tensor_projection(nlp, n, x, directions, args...)`
call the first one with zeros(n).

src/nlp/api.jl
```diff
@@ -1298,6 +1299,41 @@ function hess_op!(
...
+function tensor_projection(
+    nlp::AbstractNLPModel{T, S},
+    n::Int,
+    x::AbstractVector,
+    directions::Tuple{Int, Vararg{Int}},
+    args...
+) where {T, S}
+    @lencheck nlp.meta.nvar x
+    m = n - length(directions)
+    @assert m ≥ 1
```
tmigot, line 1324: Can we use NLPModels.@rangecheck instead of @assert ?

tmigot (changes requested):
Thanks guys for starting this! This is a first batch of comments.

src/nlp/api.jl
amontoison: Can we use `rangecheck` to ensure inequalities?

src/nlp/api.jl
amontoison: For now, we try to make it working with ForwardDiff.jl and Enzyme.jl and check that we can do that with AD such that the feature is relevant.
For Enzyme.jl, we need to wait next month and the help of Valentin but it should work for ForwardDiff.jl with multiple "forward over forward".

I will add some tests in `NLPModelsTest.jl` next week if it works as expected with `ForwardDiff.jl`.

src/nlp/api.jl
amontoison: With @michel2323, we also wanted to use the value of `n` / `order` as a type with `Val{...}` such we can rely on multiple dispatch for any tensor projection (`1 <= N < n`).

src/nlp/api.jl
amontoison: I agree that it will be great to have tensor products with both the objective and the Lagrangian.
Should we also add an API for just the constraints, or include a scaling factor `\alpha` that can go in front of the objective?

tmigot: I think it makes sense to be coherent with the other functions like `hprod`.

src/nlp/api.jl
tmigot: Not tested, but I would expect `@rangecheck 1 typemax(Int) m` to work

src/nlp/api.jl
tmigot: Looks good, thanks.

amontoison:
We never got a funding for working more on this project. I close the PR.

PR #490, opened 2025-01-27
Improve the documentation of nnzh

amontoison:


docs/src/index.md
```diff
@@ -109,7 +109,7 @@ Attribute   | Type               | Notes
 `nnzj`      | `Int             ` | number of nonzeros in the sparse Jacobian
 `lin_nnzj`  | `Int             ` | number of nonzeros in the sparse linear constraints Jacobian
 `nln_nnzj`  | `Int             ` | number of nonzeros in the sparse nonlinear constraints Jacobian
-`nnzh`      | `Int             ` | number of nonzeros in the sparse Hessian
+`nnzh`      | `Int             ` | number of nonzeros in the sparse Hessian of the Lagrangian
```
tmigot, line 112: It's actually even, the number of nonzeros in the lower triangle of the sparse Hessian of the Lagrangian :)

tmigot (changes requested):

tmigot:
Thanks @CalebDerrickson and @amontoison for reporting this! I am afraid this should be a bit everywhere in the documentation, but let's start here.

docs/src/index.md
amontoison: You're right and I forgot to talk about this triangle when I explained it to Caleb!

amontoison, line 112:
```suggestion
`nnzh`      | `Int             ` | number of nonzeros in the lower triangular part of the sparse Hessian of the Lagrangian
```

tmigot (approved):
Thanks!

PR #492, opened 2025-04-16
Improve cons! and jprod! implementation

### LINKED ISSUE #491: Significant slowdown observed in QuadraticModels.jl
I observe a significant slowdown with the current version of NLPModels. A MWE is: 

```julia
using QPSReader
using QuadraticModels
using NLPModels

src = fetch_mm()
qpdat = readqps(joinpath(src, "UBH1.SIF"))
qp = QuadraticModel(qpdat)

m = NLPModels.get_ncon(qp)
x = NLPModels.get_x0(qp)
c = zeros(m)

# Evaluate in a few milliseconds 
@time NLPModels.cons_lin!(qp, x, c)
# Computation does not complete in reasonable time
@time NLPModels.cons!(qp, x, c)
```

This second MWE is giving the behavior I am expecting: 
```julia 
using QPSReader
using QuadraticModels
using NLPModels

function NLPModels.cons!(qp::QuadraticModel{T}, x::Vector{T}, c::Vector{T}) where T
    return NLPModels.cons_lin!(qp, x, c)
end

src = fetch_mm()
qpdat = readqps(joinpath(src, "UBH1.SIF"))
qp = QuadraticModel(qpdat)

m = NLPModels.get_ncon(qp)
x = NLPModels.get_x0(qp)
c = zeros(m)

# Evaluate in few milliseconds
@time NLPModels.cons_lin!(qp, x, c)
# Evaluate also in few milliseconds
@time NLPModels.cons!(qp, x, c)
```

I am opening the issue here as I think the culprit is the following dispatch in NLPModels: 
https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/src/nlp/api.jl#L61

It returns a view, and then Julia dispatches on `generic_matvecmul!` afterwards, as the view is non-contiguous. Resulting in a significant slowdown, as we cannot use an efficient implementation for the sparse-matrix-vector product. 

What would be a valid solution? Should we always redefine `cons!` manually, or should we fix the issue in NLPModels?

tmigot:
close #491

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
+      - uses: actions/cache@v4
         env:
           cache-name: cache-artifacts
         with:
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -58,8 +58,20 @@ function cons!(nlp::AbstractNLPModel, x::AbstractVector, cx::AbstractVector)
   @lencheck nlp.meta.nvar x
   @lencheck nlp.meta.ncon cx
   increment!(nlp, :neval_cons)
-  nlp.meta.nlin > 0 && cons_lin!(nlp, x, view(cx, nlp.meta.lin))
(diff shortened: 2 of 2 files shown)


tmigot:
@frapac

amontoison (approved):

PR #519, opened 2025-11-20
Version 0.22.0

tmigot:
Release notes:

## Breaking changes
- drop support for Julia 1.6
- objcons! with in-place residual not defined for NLS
- the jac_lin* functions no longer take an x argument.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NLPModels"
 uuid = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
-version = "0.21.5"
+version = "0.22.0"
 
 [deps]
 FastClosures = "9aa1b823-49e4-5ca5-8b0f-3971ec8bab6a"


amontoison:
@tmigot You deprecated the old API or is it full breaking?
Honestly we didn't need to do this breaking change...
It is a lot of work to update all packages and it is not critical.
You should have asked before doing a breaking release...

amontoison:
@tmigot @dpo
Who will update all the packages that depends on `NLPModels.jl` in the JSO organization now?

We can't do a breaking release for just an argument in the linear API...
So much work to update all packages and do new releases of dependent packages.

amontoison:
Yes, but it will take weeks or months until release 0.22 lands in all packages.
If we yank release 0.22, we remove it from the registry, and we can fix the current branch `main` so it's no longer breaking, and continue with version `0.21.6` instead of `0.22.1`.
Releasing `0.22.1` will not fix anything, we will still need to update the compatibility entries of 30-40 packages and make new releases for all of them.

amontoison:
No Dominique, we can't have new features of minor modifications until we update the packages for 0.22.
We already got this issue in the past. I really don't want that it happens again.

amontoison:
It is not an issue if we yank the release 0.22 in the general registry, Dominique :-)
We do that quite often for the JLL.
We can add new commits on `main` to make the previous ones  non-breaking and then continue the series of release `0.21.x`.

If one day we really need a breaking release, we will tag `0.22.1`.

PR #521, opened 2025-11-20
Batch API

klamike:
to be continued in https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/540...

### DIFF
--- a/test/nlp/dummy-model.jl
+++ b/test/nlp/dummy-model.jl
@@ -13,7 +13,7 @@ end
   @test_throws(MethodError, grad!(model, [0.0], [1.0]))
   @test_throws(MethodError, cons_lin!(model, [0.0], [1.0]))
   @test_throws(MethodError, cons_nln!(model, [0.0], [1.0]))
-  @test_throws(MethodError, jac_lin_coord!(model, [1.0]))
+  @test_throws(MethodError, jac_lin_coord!(model, [0.0], [1.0]))
   @test_throws(MethodError, jac_nln_coord!(model, [0.0], [1.0]))
   @test_throws(MethodError, jth_con(model, [0.0], 1))
   @test_throws(MethodError, jth_congrad(model, [0.0], 1))


klamike:
@tmigot The idea is to define an API to talk about batches of models in a standardized way across JSO. Then each solver (up for discussion -- maybe we should do that here instead?) can define its own batch model types, which would mark the relevant assumptions (e.g. same jac/hess structure, same nvar/ncon, etc.) to enable exploiting that shared structure (importantly, not just in the evaluation of the model, but also in the solvers themselves). For example, if we assume the same jac/hess structure, we can use CUDSS uniform batch API for solving KKT systems instead of looping over single solves. This should bring substantial speedup over something like `madnlp.(nlps::Vector{<:AbstractNLPModel})`.

To be honest, this kind of change is very big for me to think about all the consequences at once. So, we decided it might be best to just implement something dumb first to get some visibility into tradeoffs.

test/nlp/batch_api.jl
```diff
@@ -0,0 +1,562 @@
...
+      # Test batch_hprod! with obj_weights (with y)
+      hprods = [zeros(n) for _ = 1:n_models]
+      batch_hprod!(bnlp, xs, ys, vs, hprods; obj_weights = obj_weights)
+      manual_hprods = [
+        hprod!(models[i], xs[i], ys[i], vs[i], zeros(n); obj_weight = obj_weights[i]) for
+        i = 1:n_models
+      ]
+      @test hprods ≈ manual_hprods
+
+      if isa(bnlp, ForEachBatchNLPModel) # NOTE: excluding InplaceBatchNLPModel
```
klamike: Had to exclude the inplace version here since the closures in the operators reference the model state, which we are mutating without their knowledge

klamike:
@amontoison I've rebased, tests should pass now. Curious what you think of this approach, as well as the downstream https://github.com/klamike/MadIPM.jl/pull/1 and https://github.com/klamike/MadNLP.jl/pull/1 (which so far do not depend on this PR)

amontoison:
@klamike @andrewrosemberg 
The PR is quite big, and we can probably make some simplifications, but it is excellent for prototyping!
I have a few comments:

* Since we are back to 0.21.x ( :muscle: ), we can try it on all models like `AMPLModel`, `CUTEstModel`, `MathOptNLPModel`, or `ExaModel`.
Can you try to run a batch of 5 ExaModels (CPU and GPU) using the recent parametric support you added in ExaModels.jl? Doing a DC-OPF with parametric QPLOAD / PLOAD could be a very nice proof-of-concept.
Andrew, maybe this task is more relevant for you?

* A `BatchNLPModelMeta` could be useful to specify what is in common or not. Inside, we could specify if we have the same `lcon, ucon, lvar, uvar, obj, ...` and avoid duplicating the arrays if not needed. This would allow a less redundant implementation.

* Another important feature is the ability to evaluate only the obj / grad / jac / ... of a specific model in the batch. We should provide a variant of the function where we can specify an index of the problem, a subset of the problems, or all problems.

We don’t want to update all KKT systems if some of them have already converged.

PR #522, opened 2025-11-21
Implement default jac_lin* API to patch 0.22

tmigot:
A suggestion of changes to make 0.22 less breaking.
I tested NLPModelsModifiers, NLPModelsTest, QuadraticModels, ADNLPModels .

Changes are very minor, for instance for NLPModelsModifiers and NLPModelsTest:
- Testing a poor x would no longer fail
- The show function of nlp has also changed

QuadraticModels and ADNLPModels were not requiring any modification.

I commented out the warn messages in some places, because it allocates.

### DIFF
--- a/src/nlp/api.jl
+++ b/src/nlp/api.jl
@@ -316,10 +316,10 @@ overwriting `vals`.
 """
 function jac_lin_coord! end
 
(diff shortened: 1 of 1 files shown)


dpo:
Presumably, they test before merging. In the mean time, let's fix the issue. I'm also willing to help upgrading the various dependents here. I'm by no means asking you to do it all by yourself.

@tmigot Says that he tested this PR against several packages. Now I'd like to understand why the breakage isn't working.

PR #525, opened 2025-12-01
Add availability flags in NLPModelMeta and NLSMeta

amontoison:
Add flags in `NLPModelsMeta` and `NLSMeta` to specify whether gradient, sparse Jacobians, sparse Hessians, and operator-based products are available in a model.
In some models, we do not want to, or cannot, implement the complete `NLPModels.jl` API.

Examples:

* `ADNLPModels.jl`: we do not want to set up some AD backends if they are not needed (see issue https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/360).
* `NLPModelsJuMP.jl`: the user can specify from JuMP which subset of derivatives is needed, and the new `VectorNonlinearOracle` structure in MOI does not support operator–vector products.
* Custom `AbstractNLPModel` implementations: at Argonne, we have some models involving neural networks where only the gradient is available (cc Sarah).

This is an issue for solvers, because solvers such as `MadNLP.jl` or `MadNCL.jl` expect `jtprod` to be implemented but cannot easily know whether it is available before calling it.
A similar issue occurs with `UnoSolver.jl`, which relies on the BQPD subsolver by default and requires `hprod`.
The absence of the Lagrangian Hessian can also help solvers like `NLPModelsIpopt.jl` or `NLPModelsKnitro.jl` to automatically switch to quasi-Newton approximations.

Using these new attributes also helps an oracle choose the most appropriate solver, and ensures that a clean error is returned when a solver cannot be used with a given model (`JSOSuite.jl`?).
This is preferable to triggering a missing method error.

This addition should be non-breaking (the full API is considered available by default) and should resolve a number of issues in dependent packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,40 +84,46 @@ The complete list of methods that an interface may implement can be found in the
 
 `NLPModelMeta` objects have the following attributes (with `S <: AbstractVector`):
 
-Attribute   | Type               | Notes
-------------|--------------------|------------------------------------
-`nvar`      | `Int             ` | number of variables
-`x0  `      | `S`                | initial guess
-`lvar`      | `S`                | vector of lower bounds
-`uvar`      | `S`                | vector of upper bounds
-`ifix`      | `Vector{Int}`      | indices of fixed variables
-`ilow`      | `Vector{Int}`      | indices of variables with lower bound only
-`iupp`      | `Vector{Int}`      | indices of variables with upper bound only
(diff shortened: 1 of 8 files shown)


amontoison:
@tmigot The motivation is here: https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/524#issue-3679588617

PR #526, opened 2025-12-03
Remove ADNLPModels.jl from docs/Project.toml

### LINKED ISSUE #520: Remove ADNPLModels from NLPModels.jl documentation
Analyse what would be different.
The main issue is that whenever you make a breaking release for NLPModels it breaks the doc.

IMO, it's fine to remove the examples here https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/docs/src/tools.md .

amontoison:
close #520

### DIFF
--- a/docs/src/tools.md
+++ b/docs/src/tools.md
@@ -6,7 +6,7 @@ After calling one the [API](@ref) functions to get a function value, the
 number of times that function was called is stored inside the
 `NLPModel`. For instance
 
-```@example
+```julia
 using ADNLPModels, LinearAlgebra, NLPModels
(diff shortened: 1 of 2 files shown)


docs/src/tools.md
```diff
@@ -6,7 +6,7 @@ After calling one the [API](@ref) functions to get a function value, the
 number of times that function was called is stored inside the
 `NLPModel`. For instance
 
-```@example
+```julia
```
tmigot: At this point, I think it's fair to just remove the two examples.

PR #540, opened 2026-02-04
Batch API

amontoison:
cc @klamike

### DIFF
--- a/src/nlp/batch_api.jl
+++ b/src/nlp/batch_api.jl
@@ -17,6 +17,11 @@ function obj(bnlp::AbstractBatchNLPModel{T, S}, bx::AbstractMatrix) where {T, S}
   return bf
 end
 
+"""
+    bf = obj!(bnlp, bx)
+"""
+function obj! end
+
 function grad(bnlp::AbstractBatchNLPModel{T, S}, bx::AbstractMatrix) where {T, S}
   bg = S(undef, bnlp.meta.nvar, bnlp.meta.nbatch)
   grad!(bnlp, bx, bg)


amontoison:
Michael, I finished what I wanted.
You can do a pass on the tests when you have time.

amontoison:
Amazing Michael!

klamike:
Yes I meant updating the `nbatch` throughout the solve. At the NLPModels level it would just change the size of the buffers in the out-of-place API. The motivation is to have some way of skipping evaluating batch elements that have already converged, for batch models where that can be done efficiently. Models with non-independent elements can override it to error of course. Similar story for the `get_nlpmodel_at_index!`.

In the ExaModels case, of course it depends on how you do the batching. When I added parameters to ExaModels I specifically made the lower level functions all take the parameter vector as an input, to make it possible to implement the batching the way I did in [BatchNLPKernels](https://github.com/LearningToOptimize/BatchNLPKernels.jl). It is based on a single ExaModel and does the same number of kernel launches for the batch evaluation as ExaModels would for a regular model, just with 2D grids. Since the base parametric ExaModel is built "unbatched", it is trivial to implement the `set_nbatch` and `get_nlpmodel_at_index!`.

amontoison:
> Sorry about that, I missed your message.. The latest result is on a batch of 128 9241_pegase DCOPF, ~6.5x faster over sequential, and ~1.85x faster over multi-threading with 8 threads (comparing 1 task/mini-batch vs 1 task/problem)

Good :1st_place_medal: 
For MadIPM.jl, should we provide `jfix`, `jlow`, ... as a boolean vector of size `ncon * nbatch` instead of only local indices?
It hard to get the coefficient for each batch with the current version.

Also should I add a prefix `batch_` for the availability flags in `AbstractBatchNLPModelMeta` ?

@michel2323 worked on `BatchExaModel` today, and except the function `batch_obj`, we just need to rename the routines to follow this batch API.
--> https://github.com/exanauts/ExaModels.jl/pull/216

sshin23:
Which GPU did you run the benchmark on @klamike?

klamike:
I believe it was RTX 6000 Pro Blackwell

klamike:
We are indeed reusing the symbolic factorization, using cuDSS's uniform batch feature. But the overall performance of the solver is not quite optimized, currently a lot of time is spent on packing/unpacking from individual to batched buffers since that made it simpler to batch-ify incrementally. Now that (most of) the solver is batched, I plan to revisit this.

I think we have H100 and H200 but no B100. I'll give it a try soon, with more CPU threads too.

klamike:
I like that approach! But for the KKT system nzval, I think strided vector is actually nicer (we can reuse the transfer kernels by just building a batched map). Matrix is definitely more natural for a user-facing API. (so it makes sense to have both in CUDSS).

As I am working on this version, I have come across several kernels which can be written exactly the same (differs only in argument types), or only one or two words different (e.g. same `mapreduce` but with a `dims=1`), between batched and unbatched. I'm trying to make it match as closely as possible so we can eventually just have one kernel, shared between batched and unbatched. For that, keeping things as Matrix is very helpful. The name change you made from `batch_obj` -> `obj` is nice for that too.

klamike:
Actually you are right, Matrix is nice even for the nzval... I am switching it :)

amontoison:
I will update `BatchNLPModelMeta` tomorrow such that `lvar`, `uvar`, `lcon`, `ucon` are matrices.

PR #557, opened 2026-02-28
Parametric AD API

### LINKED ISSUE #532: Parametric API
It would be nice for NLPModels to define a standard API for the data we'd need for sensitivity analysis. In particular:

- Partial second derivatives of the lagrangian w.r.t. primal solution and parameters (e.g. `NLPModels.hess_param(nlp, x, y)`)
- Partial derivatives of the constraints w.r.t. parameters (e.g. `NLPModels.jac_param(nlp, x)`?)

klamike:
This is needed to support efficient implicit differentiation, e.g. https://github.com/MadNLP/MadDiff.jl.

The idea is to have parameters be fully "implicit" in the eyes of NLPModels, instead of lifting them to "special variables" like in JuMP/DiffOpt. While the latter approach allows to reuse the existing AD APIs (slicing out the parametric parts), this approach should be more efficient. It also allows for variable/constraint bounds to be parametric.

cc @amontoison 

closes #532


src/nlp/param_api.jl
```diff
@@ -0,0 +1,534 @@
...
+function grad_param! end
+
+"""
+    (rows, cols) = jac_param_structure(nlp)
+
+Return the structure of the constraints Jacobian wrt parameters in sparse coordinate format.
+This function is only available if `nlp.meta.jac_param_available` is set to `true`.
+"""
+function jac_param_structure(nlp::AbstractNLPModel)
+  rows = Vector{Int}(undef, nlp.meta.nnzjp)
```
apozharski: Here and elsewhere I would advise using  `get_*(nlp.meta)` instead of directly using field access. This will allow downstream packages to define their `*Meta` structures as they see fit rather than having the whole cavalcade of fields, most of which are not actually relevant to the model at hand.

klamike:
This is ready for review/merge now. I've added docs, tests, and an API for get/set parameter values.

klamike:
Closing in favor of the separate repo approach. @dpo let me know how you want to do it. I'm thinking you should create the repo so that it has all the template/codecov/actions set up the way you like?

PR #495, opened 2025-05-13
Optimize jac_coord! and jac_structure!

amontoison:
cc @frapac

PR #564, opened 2026-04-17
Simplify the implementation of jprod!

dpo:
I find `jprod!()` unnecessarily complicated.


amontoison:
@dpo The code is quite complicated because we got a performance regression wity the linear API (see #491).
The PRs #492 and #495 are related to this issue.

dpo:
Oh I see, thank you. It's really too bad that views cause so much trouble. I'll close this then.

amontoison:
Views could be fine as long as they are strided. If they are not strided, we can't use BLAS / LAPACK (CPU and GPU) and Julia dispatches to generic fallbacks.

